Issue #20, opened 2025-11-04
[Feature] Playing music


Tazi0:
Currently the music tab is shown in the library. The tab shows artists like they are movies, so it tries to play them as movies.

You probably want to focus first on movies and tv-shows before adding music playing (if you even want that). So I would suggest hiding the tab

Though if you were to add the feature, might I suggest showing the albums in the library (or having a switch option) and playing them like Plexamp does?

<table>
<tr>
 <td>
<img width="651" height="899" alt="Image" src="https://github.com/user-attachments/assets/7487d503-c319-44a9-92c5-c9327b327483" />
 <td>
<img width="651" height="899" alt="Image" src="https://github.com/user-attachments/assets/55be1f85-8359-46b6-9a89-6225121af5de" />
 <td>
<img width="651" height="899" alt="Image" src="https://github.com/user-attachments/assets/3ada5c02-7af5-406f-ae84-43c247753543" />
<tr>
 <td>
Music tab
 <td>
Artist page
 <td>
Trying to "play" the artist
</table>

edde746:
I will look into music and photos in the future, for now the main focus is videos. If someone else wants to beat me to it, PRs are always welcome!

PR #32, opened 2025-11-04
Hide music libraries from app (Issue #20)

TheLidlMan:
- Filter out artist-type libraries in libraries_screen.dart
- Add safety check in media_card.dart to prevent music playback attempts
- Filter music content from getOnDeck() and getRecentlyAdded() in plex_client.dart

Music libraries are now hidden throughout the app since music playback is not yet supported. This focuses the app on video content (movies and TV shows).

#20


edde746:
You're on a very old commit

TheLidlMan:
Yep, fast moving repo. Sorting now

Issue #47, opened 2025-11-05
[Feature Request] Sleep Timer


RockoWallaby:
Would be great to have a sleep timer feature.

PR #52, opened 2025-11-05
feat: sleep timer

TheLidlMan:
- default timer values in settings
- extend timer function
- pauses playback when done

#47 

Theres alot of chunky code in the `lib/widgets/plex_video_controls.dart` dunno if you want to split it out?


edde746:
Is there a reason 30 minutes is marked as "default"? By default the sleep timer should be off, right?

TheLidlMan:
Its not on just the default time when you click the button in the player so people can set custom times. Maybe it should be set to no default?

edde746:
I think it looks weird to have a "default" on the interface.

PR #73, opened 2025-11-08
feat: audiobook support

8bitgentleman:
So few plex clients have support for audiobooks so I decided to work with Claude to add one. This adds comprehensive audiobook functionality to plezy including:

- Audiobook library browsing and detail screens
- Full-featured audiobook player with chapter navigation
- Author browsing and filtering
- Audiobook-specific UI widgets and controls
- Playback state management and seeking
- Integration with existing library navigation

I don't think the main app has background play support yet so I held off on that since it seemed much more invasive


edde746:
Music/audiobook support should probably be done as one, proper OS controls is also a requirement and for some reason it's very broken with media_kit on iOS, until there is a solution for that I don't to add this as it would not support background play.

PR #79, opened 2025-11-09
feat: Adaptive navigation bar

Tazi0:
The bottom navigation bar is now adaptive for different platforms.

Using [adaptive_platform_ui](https://pub.dev/packages/adaptive_platform_ui)

Scherm­afbeelding 2025-11-09 om 11 36 58
Scherm­afbeelding 2025-11-09 om 11 38 33

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1,4 +1,6 @@
 PODS:
+  - adaptive_platform_ui (0.1.0):
+    - Flutter
   - Flutter (1.0.0)
   - media_kit_libs_ios_video (1.0.4):
     - Flutter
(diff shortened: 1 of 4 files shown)


edde746:
Seems to be quite buggy, it will randomly become small when scrolling down.
Simulator Screenshot - iPhone Air - 2025-11-09 at 16 27 59

Also, while I personally like the liquid glass design, it doesn't really fit with the rest of the app.

Tazi0:
Thanks for reporting the bug, I will look into it. The navigation bar is not the only thing that can be adaptive I can also edit the other elements but that would take significant time which I'm not sure is a great usage of time.

edde746:
Yeah, the issue however with the [adaptive_platform_ui](https://pub.dev/packages/adaptive_platform_ui) package is that it uses AppKit instead of SwiftUI, which means there isn't macOS support which would be nice to have.

Issue #43, opened 2025-11-04
Feature request: Skip intro/credits


Seros:
The button to skip intros and credits on demand like in the official apps would be great. The docs mention that this requires the server admin to subscribe for a Plexpass but not sure if this is required for the clients as well and can be read via an api.

TheLidlMan:
Yes, its possible and already 'kinda' implemented with the chapters. Its only possible if the server admin has Plex Pass as you said. The marker is in the file so once its been recognised then should be readable by client, possibly a future feature

Tazi0:
I created a merge request to add this functionality

PR #80, opened 2025-11-09
feat: Skip intro and credit

Tazi0:
A feature request from [Seros](https://github.com/Seros) #43.

The buttons are shown based on chapters. 

> [!NOTE]
> Only issue is that the final chapter doesn't have to be on the credits. The series I tested has a "next episode" preview so the credits are actually the chapter before the last. Any suggestions?

<table>
<tr>
 <td>
Scherm­afbeelding 2025-11-09 om 13 15 19
 <td>
Scherm­afbeelding 2025-11-09 om 13 15 12
<tr>
<td>
Scherm­afbeelding 2025-11-09 om 13 22 00
<td>
Scherm­afbeelding 2025-11-09 om 13 22 54
<tr>
 <td>
Skip Intro
 <td>
Next episode
</table>

### DIFF
--- a/lib/client/plex_client.dart
+++ b/lib/client/plex_client.dart
@@ -501,18 +501,17 @@ class PlexClient {
   Future<List<PlexMarker>> getMarkers(String ratingKey) async {
     final response = await _dio.get(
       '/library/metadata/$ratingKey',
-      queryParameters: {'includeChapters': 1},
+      queryParameters: {'includeMarkers': 1},
     );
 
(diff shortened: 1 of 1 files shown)


edde746:
Plex does provide this information when you have Plex Pass (I think), we should rely on that instead of custom logic using chapters.

Tazi0:
well I'll be dammed, `includeMarkers` does exist. It wasn't documented

Issue #86, opened 2025-11-10
Subtitles cut off in Letterbox mode


lhilden:
Expanding video to Letterbox mode cuts off subtitles

### Expected
All subtitles are visible. Example:
<img width="2048" height="858" alt="Image" src="https://github.com/user-attachments/assets/69eda8f4-76c4-45c3-b1c7-035bb77c4517" />

### Actual
Subtitles are cut off. Example:
<img width="2048" height="858" alt="Image" src="https://github.com/user-attachments/assets/a67a9519-bbe7-4b50-abc5-57e0154ca584" />

### Affects Versions
Platform: macOS
Plezy Version: 1.6.1

Tazi0:
I created a merge request to fix this.

edde746:
Fixed with https://github.com/edde746/plezy/commit/0b9998ddffb3090561abd4594857fb0bc5f5f122

PR #89, opened 2025-11-11
feat: Custom subtitles overlay

Tazi0:
A bug report from [lhilden](https://github.com/lhilden) #86.

This fixes the changing of aspect ratio for the video (Letterbox -> Fill screen -> Stretch). 

A custom subtitles overlay which is always on the bottom, but bottom distance dependent on the video control overlay and device.

<table>
<tr>
 <td>
Scherm­afbeelding 2025-11-11 om 12 47 42
 <td>
Scherm­afbeelding 2025-11-11 om 12 47 38
<tr>
 <td>
Scherm­afbeelding 2025-11-11 om 12 48 39
 <td>
Scherm­afbeelding 2025-11-11 om 12 48 42
<tr>
 <td>
With video control overlay
 <td>
Without video control overlay
</table>

### DIFF
--- a/lib/screens/video_player_screen.dart
+++ b/lib/screens/video_player_screen.dart
@@ -68,7 +68,10 @@ class VideoPlayerScreenState extends State<VideoPlayerScreen> {
   int _boxFitMode = 0;
   bool _isPinching = false; // Track if a pinch gesture is occurring
 
-  // Controls visibility state for subtitle positioning
+  // Subtitle overlay state
+  bool _showSubtitleOverlay = false;
+  String _currentSubtitleText = '';
+  Timer? _subtitleTimer;
   bool _controlsVisible = true;
 
   @override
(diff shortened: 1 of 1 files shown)


wanestar:
Wouldn’t this break .ASS subtitles with styling and positioning?
Feels like there could be a better way than completely bypassing MPV, unless I’m reading it wrong.

At the very least there has to be a way to disable this. I really like MPVs native subtitle rendering.

edde746:
I assume @wanestar is right, I was looking at how https://github.com/DonutWare/Fladder does it and I think this is the same solution. SSA support is a must, the proper solution is to move away from using BoxFit to implement the fill mode and instead using mpv properties.

Tazi0:
I wasn't aware of .ASS subtitles but your right. I'm currently trying to work subtitle margining into the video player. But that's quite difficult to manage the bottom position in different with the boxfits and devices.

Tazi0:
I'm currently not finding any way to have three different fit sizes (fit, cover, stretch) and keep the subtitles in the frame of the screen apart from the solution i've given before. If anyone else wants to stab at this.

edde746:
The solution is to use the following mpv properties:
```
'sub-margin-y': '300',
'sub-ass-force-margins': 'yes',
'sub-ass-style-overrides': 'MarginV=300',
```
(the 300 value is just an example, we need to actually calculate the right value)
but, the mpv version that we're using on iOS & macOS (0.36.0) does not support this, it needs to be updated.

edde746:
See #86

PR #102, opened 2025-11-14
add zh i18n

ulovecode:


edde746:
@ulovecode Are you able to check the suggestions Copilot made?

PR #125, opened 2025-11-21
Add unit tests

Doezer:


Doezer:
@edde746 hey, can you check this PR? Whether or not it's accepted :)

edde746:
The tests aren't very useful, I don't really know what could be unit tested in a useful way. End-to-end tests would be more useful but hard to do since we're relying on Plex APIs, there is some work done on this in the `maestro/` folder.

Doezer:
Sure, I understand; I'll close the PR so I can continue working on main

PR #133, opened 2025-11-26
fix:  save stream selection per media for audio and subtitles

Doezer:
There is currently a bug (or maybe more of a gap) that doesn't allow saving the subtitle selection for media. With this change, Plezy now saves the user selection to Plex when they change the subtitle.
I didn't remove the existing behaviour of saving the language to the series (for tv shows), but it might be redundant for movies.

I'm working on a third PR to fix the issue of not actually using the subtitle track sent by Plex as the selected one.

### DIFF
--- a/lib/client/plex_client.dart
+++ b/lib/client/plex_client.dart
@@ -493,6 +493,9 @@ class PlexClient {
 
         return response.statusCode == 200;
       }
+      // Si allParts est false, retourner true ou false explicitement (selon la logique souhaitée)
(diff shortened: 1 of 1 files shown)


edde746:
I don't think the hardcoded `languageMap` is needed, we already have language code data in a json file. If that data is not the one needed, please add another json with all the required language codes instead of hardcoding the map.

Doezer:
@edde746 I updated the PR to use the existing map. Also rebased.

Issue #135, opened 2025-11-27
Adaptive icon is too big on Android


mkaflowski:
### Bug Description

<img width="207" height="119" alt="Image" src="https://github.com/user-attachments/assets/1151f97f-3c9c-4918-8d7b-437ac4ac92c1" />

Adaptive icon is too big.

### Steps to Reproduce

Change icon according to https://developer.android.com/develop/ui/views/launch/icon_design_adaptive
You can import icons with File-New-Image Asset

### Expected Behavior

Icon size matching to the rest.

https://developer.android.com/static/guide/practices/ui_guidelines/images/adaptive-icon-themed-icons.png

### Actual Behavior
Icon is too big.

### Plezy Version

1.8.1

### Platform

Android

### Device Information

Pixel 8

edde746:
Android icon system is so shit 🥀

PR #136, opened 2025-11-27
Android monochrome icons size fixed

mkaflowski:
Fix for https://github.com/edde746/plezy/issues/135#issuecomment-3586851649


edde746:
I squashed and merged this, but because of the squash GitHub doesn't recognize it as merged.

PR #137, opened 2025-11-29
fix: Get builds running on iOS simulator

JackDanger:
Prior to this change the linker would error on x86 architectural problems
If we're building on an ARM Mac this allows simulator build and boot.

And thank you _so_ much for making Plezy!


edde746:
Closing this as we are no longer using media_kit.

PR #187, opened 2025-12-23
Optimize MediaCard widget by caching computed values 

Doezer:


edde746:
Do you have any performance measurements before and after? It doesn't seem to be worth adding the complexity unless there is actually a performance issue.

edde746:
Here are the results from ~800 widget builds (scrolling through a library):

  | Metric                      | Caching ON | Caching OFF | Difference  |
  |-----------------------------|------------|-------------|-------------|
  | _buildSemanticLabel() total | 4,248µs    | 4,800µs     | 552µs saved |
  | _getLocalPosterPath() total | 306µs      | 374µs       | 68µs saved  |
  | Total                       |            |             | ~620µs      |

  The caching saves approximately 0.6ms total across an entire browsing session with heavy scrolling, less than 1 microsecond per build.

  Given that the optimization:
  - Adds ~30 lines of lifecycle management code (initState, didChangeDependencies, didUpdateWidget)
  - Increases maintenance complexity
  - Could introduce subtle bugs if item data is ever mutated in-place
  - Provides sub-millisecond benefit across hundreds of builds

I don't think the tradeoff is worthwhile.

Doezer:
Thanks for checking :) I didn't have access to my computer right away

Issue #44, opened 2025-11-05
[request] picture-in-picture for background play


twilsonco:
When Plezy is minimized while video is playing, it should continue playing using picture-in-picture. 

(Plezy rocks)

arennow:
I would also like an explicit button for it – I find that easier (and more discoverable) sometimes than minimization. PIP is a big part of my "workflow" in watching stuff (on macOS), so I'll be very excited when this gets implemented :)

andperse:
This is what's keeping me from using plezy on mobile (android) instead of the official client.

Android has a setting to enable picture in picture, but apparently the app must have the capability turned on.

dansl:
If this isn't actively being worked on yet, I will try forking the code and put in a pull request. 

I use PiP all the time, and the official Plex app will cause my phone to hang for about 5-6 seconds when I activate PiP... so having this in Plezy would be great!

edde746:
Not currently being worked on, best of luck.

nphalem:
@edde746 it's been added to android at least. Wanna close it or do you leave it open for other platforms?

RileyStoned:
Would love this to be added to the IOS app and Mac OS app too, I think this is the only feature missing I want before moving over completely.

svenjr:
I made a branch locally with changes for MacOS but don't have permission to push a new branch to this repo 😢

dansl:
> I made a branch locally with changes for MacOS but don't have permission to push a new branch to this repo 😢 

Make a "pull request", and if your code passes inspection, it will get merged. You will need to fork the project and then make the pull request.

BartSiwek:
I see this is not prioritized for iOS/iPadOS/macOS. That is a shame - other players like official Plex app or Infuse have this functionality. Other non-plex related apps like YouTube, Twitch, Kick and anything that plays video has this feature. It does make the app feel unfinished in the Apple ecosysem as one of the fundamental features is missing.

edde746:
@BartSiwek It's not an easy feature to implement when using a custom video player instead of the system default one.

BartSiwek:
I agree, but it's different to say "this is not prioritized" vs "we are working on it but it will take a while".

Having said that I would love to help, however I would need significant guidance due to lack of familiarity with Flutter (Swift I somewhat know, but it has been years since I used it last).

On the PR you mention "KSPlayer, IINA, etc" - if you can point me more precisely I can at least see how much over my head this is.

[...]

PR #206, opened 2025-12-31
Add Picture-in-Picture support for Android video playback

dansl:
Implemented native Picture-in-Picture (PiP) mode for Android by updating the AndroidManifest, MainActivity, and introducing PipService and VideoPIPManager. Updated video player screen and controls to support toggling PiP mode, and added localized strings for the PiP button in multiple languages.

Relating to Issue #44 : At this time PiP in Plezy only supports Android devices... iOS+MacOS will require significant changes to the video player for it to work properly...


andperse:
@dansl Picture in picture on android mobile only triggers when I press the dedicated button in the player, is this intentional? If so, I'll submit a request for it to trigger when switching to another app or the home screen.

Issue #201, opened 2025-12-28
Windows: livestream to discord show black sceen on player


mus6677:
### Bug Description

Since 1.9 release, choose to live stream plezy on discord only show black on the video player. 

### Steps to Reproduce

- join dicord voice channel
- start live stream by choosing plezy windows application in discord
- play any video in plezy
- the shared livestream only show black on video player

### Expected Behavior

if revert back to 1.8.1, dicord can show the video playback no problem

### Actual Behavior

black screen on video player

### Plezy Version

1.13

### Platform

Windows

### Device Information

PC, nvidia gpu

### Logs and Stack Traces

```shell

```

### Screenshots

<img width="2197" height="1271" alt="Image" src="https://github.com/user-attachments/assets/17a07ddb-2b41-4536-97ed-0eaba20b832d" />

edde746:
~~Workaround merged, for anyone looking at this in the future, check my comment on the PR.~~

Workaround had issues when toggling fullscreen. Screen sharing remains remains broken.

germanletsnick:
Funny thing, if you are streaming your whole desktop its working totally fine. Only if u select to stream Plezy specifically its showing black

sekhu-net:
same issue. selecting ONLY the app in Discord causes  a black screen. Unfortunate, I like the app otherwise

sekhu-net:
Just an FYI that I tried the latest release and still experience the same issue

Blizzone191:
Same here still having same issue as well. It's like when trying to stream Netflix everything shows until you play a video

edde746:
Some information about this:

It can't easily be fixed, this is due to how the Plezy GUI and mpv is composed on Windows. The "easy" "solution" would break HDR on Windows, which I don't think is a worthwhile tradeoff.

The proper solution is one of the following:
- Engine fix, either wait for Flutter to natively support HDR or create an engine patch
- Custom embedder, reimplementing the Flutter embedder to better compose mpv

Both of these are massive amounts of work and would pretty much only be to fix this bug (and be technically cleaner/more proper).

There is no need to post here unless you have information that would help solve this.

PR #229, opened 2026-01-11
Workaround for Discord Screensharing on Windows

ZhunCn:
Since the mpv player implementation in version 1.9.0, Discord screenshares/live streams on Windows machines have been appearing to have black video screens.

These changes (after so much back in forth with Gemini 3 Pro and filtering out non-essential changes) is the minimum workaround to allow Discord to get window capture while still allowing normal playback on the client (some changes suggested by Gemini would cause the video stream to go transparent or black or is completely unnecessary to get the end result).

Note that this creates a separate "window instance" that the discord user needs to select. The original window instance still shows up as a black video stream, which is why I call it a workaround than an actual fix.

image

This is one of my first open source PR, so please give me any feedback as needed and adjust the PR title.

Related Issue: #201

### DIFF
--- a/windows/runner/mpv/mpv_container.cpp
+++ b/windows/runner/mpv/mpv_container.cpp
@@ -30,12 +30,14 @@ HWND MpvContainer::Create() {
   ::RegisterClassExW(&window_class);
 
   // Use WS_POPUP for a borderless window without title bar.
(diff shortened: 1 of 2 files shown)


edde746:
Works as expected, nice. I think the only "proper" solution while maintaining HDR support would be to make a custom Flutter embedder. One issue with this is that it seems to make the stream completely white when no video is played, effectively flashbanging any viewers.

PR #297, opened 2026-01-19
feat: Open libraries on Recommended tab when selected via sidebar

dwk001:
- Force tab 0 on library selection from side nav
- Persist choice for consistency; still restores saved tab when navigating within Libraries


edde746:
Why?

dwk001:
Often I'll open (at least on Android TV) a library and the tab at the top will default to something other than recommended. This creates a less than ideal user experience. The recommended tab should simply be the default behavior since it will definitely show content.

edde746:
It will default to the one you were on last time you were in that library.

dwk001:
I think this is a side effect of going into library management. Because on TV you have to scroll through your tabs to get to library management pencil. I've considered an enhancement where when you long press or right click on Libraries it allows you to click and access manage libraries without having to scroll over. 

Let me know what you think I can work on that instead if you'd be open to it

edde746:
That would be a better solution

Issue #232, opened 2026-01-14
Enable play/pause by clicking anywhere on the video player


germanletsnick:
### Problem Statement

Hi there! Thank you for maintaining this project.

I’m currently using the player and noticed that play/pause functionality is currently restricted to the button in the bottom-left corner. I would like to suggest adding the ability to start, pause, and resume the video by clicking anywhere on the video frame itself.

Why this would be helpful:

Efficiency: It’s much faster to click the center of the screen than to aim for a small button in the corner.

Standardization: Most modern web players (like YouTube, VLC, or Netflix) use this behavior, so it makes the player feel more intuitive for users.

### Proposed Solution

Single Left-Click: Toggle between Play and Pause.

Existing Controls: Keep the bottom-left button as a secondary option for those who prefer it.

Thank you for considering this improvement!

### Target Platform

Windows

edde746:
Use the spacebar, clicking will be kept for toggling controls.

germanletsnick:
Could you maybe add a Toggle in the Settings, where you can choose what happens if you click on the video please?
Or maybe a big play Button in the middle of the screen? like on android

lhilden:
@edde746  
This is a totally reasonable request to make Plezy semantically compatible with other video players. I really hope you reconsider implementing it. We're all assuming that clicking on the video will pause it.

This project is off to a great start, It would be sad to squander the goodwill earned on this project thus far by vetoing simple features. Especially when its just asking Gemini or Claude to generate it.

edde746:
@lhilden You're not entitled to make design decisions. If you don't like it, maintain a fork.

lhilden:
@edde746 so much for goodwill.

micahmo:
Please reconsider this request! It's very unusual for a desktop video player to not play/pause when clicking on the video. Keeping this behavior will make Plezy confusing for those who are coming from pretty much any other type of desktop video player.

If you personally don't like the behavior, then it can be an option, off by default.

Would you accept a PR with this functionality?

edde746:
Depends on how good the PR is.

PR #305, opened 2026-01-20
Add the ability to click the video player to play/pause

micahmo:
For #232, this PR adds the ability to play/pause video playback by tapping anywhere on the video player (instead of needing to tap specifically on the play/pause button) when running on a desktop. (On mobile, tapping still toggles the overlay.) This brings Plezy in line with other desktop video players like the official Plex app and YouTube.

**Notes**
 - Technically `_handleControlsOverlayTap` will only fire when the user clicks the overlay (in other words, the overlay has to be visible for the play/pause toggle to work). In most cases this should be fine, because the overlay appears when the user hovers the mouse. So if they drag their mouse over the player and then click, everything should work fine. There is one scenario where it might not work exactly like how the user expects. If they leave their mouse over the video player, _and_ the video is playing, _and_ then the overlay auto-hides, _and_ then they click _without_ moving the mouse at all, the first click will technically show the overlay, and the second click the toggle the video. I don't think this behavior is too bad but I'm open to feedback.
 - The new behavior is optional and **off by default** to maintain the existing behavior for upgrading users.
 - Since the new setting only applies to the desktop, it is only shown when the app is running on desktop.

### Demo

https://github.com/user-attachments/assets/348e0625-5e83-413c-8d58-96ddba53ab03

### DIFF
--- a/lib/widgets/video_controls/video_controls.dart
+++ b/lib/widgets/video_controls/video_controls.dart
@@ -786,6 +786,15 @@ class _PlexVideoControlsState extends State<PlexVideoControls> with WindowListen
     widget.onSeekCompleted?.call(position);
   }
 
+  /// Handle tap in skip zone for desktop mode
+  void _handleTapInSkipZoneDesktop() {
+    if (widget.canControl && _clickVideoTogglesPlayback) {
+      widget.player.playOrPause();
+    }
+
+    _toggleControls();
+  }
+
   /// Handle tap in skip zone with custom double-tap detection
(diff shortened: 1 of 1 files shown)


edde746:
I think clicking when the controls are hidden should pause/unpause and show the controls.

micahmo:
> I think clicking when the controls are hidden should pause/unpause and show the controls.

That's totally fair! I will take a look at adding that.

micahmo:
> I think clicking when the controls are hidden should pause/unpause and show the controls.

@edde746 This should work now!

PR #307, opened 2026-01-20
Fullscreen Immersive button for Windows Client (accessible on top bar)

GamesDotdev:
image


edde746:
Excellent! What exactly does it do?

GamesDotdev:
> Excellent! What exactly does it do?

When I enter in fullscreen with the player i cant leave the immersive mode if i am no longer on the video but on the main menu, this way we can exit it but also enter it without being on the video player since we cant reach this button. Not sure what the effects are on other platforms this is for plezy windows :) 
image

edde746:
[Already fixed](https://github.com/edde746/plezy/commit/852f778c4244d417f203c133e19ebaccfb857207). Any other reason for this?

GamesDotdev:
ah then no sorry did not know it was already added, you can close it sorry :)

Issue #299, opened 2026-01-19
bug: Slow closing, SIGSEGV (Linux)


Nedlinin:
### Bug Description

The app loads up great but when I click the X to close the window it typically pauses for about 3-4 seconds then disappears.  Loading the app in the console shows no output while closing until it returns a SIGSEGV (Address boundary error) and the window goes away.

I've noticed that if I go in to the app settings->About menu  (or similar sub menu) and close the app it closes without issues.

### Steps to Reproduce

Open app.
Close app.

### Expected Behavior

Close should be nearly instant, no memory address errors.

### Actual Behavior

SIGSEGV

### Plezy Version

1.14.1

### Platform

Linux

### Device Information

Operating System: CachyOS Linux 
KDE Plasma Version: 6.5.5
KDE Frameworks Version: 6.22.0
Qt Version: 6.10.1
Kernel Version: 6.18.5-2-cachyos (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 5800X3D 8-Core Processor
Memory: 32 GiB of RAM (31.3 GiB usable)
Graphics Processor: NVIDIA GeForce RTX 3080 Ti

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
> plezy
Reading initial gamepads...
Listening for gamepads...
** Message: 10:25:32.905: MPRIS interface initialized successfully
Gdk-Message: 10:25:33.071: Unable to load  from the cursor theme
[I]  GamepadService: Starting on linux
[I]  GamepadService: Found 0 gamepad(s)
[I]  GamepadService: Listening for gamepad events
[I]  Refreshed 1 servers from API
[I]  Connecting to 1 servers...
[I]  Connecting to 1 servers...
[I]  Connection race found first working endpoint  ERROR: {uri: http://<REDACTED>:32400, type: Local}
[I]  Successfully connected to plex-new
[I]  Connected to 1/1 servers successfully
[I]  Starting network monitoring for all servers
[I]  Successfully connected to 1 servers
[I]  Servers connected - performing startup sync
[I]  📚 Loaded 0 episode counts from SharedPreferences
[I]  Loaded 0 downloads, 0 metadata entries, and 0 episode counts
[I]  Connectivity restored - starting bidirectional watch sync
[I]  Connection race found first working endpoint  ERROR: {uri: http://<REDACTED>:32400, type: Local}
[E]  Failed to load home users  ERROR: type 'Null' is not a subtype of type 'String' in type cast
[...]

PR #308, opened 2026-01-20
Stabilize crashes on Linux window close shutdown

Nedlinin:
Fixes https://github.com/edde746/plezy/issues/299

Admittedly didn't test this on anything other than Linux (don't have access to another box currently).

The gist is that we now listen for the close event and blank out the widget tree.  Additionally we do some clean ups for the video player itself during closing.


edde746:
"Merged" the first commit since the second one it just merging main into your branch.

edde746:
Actually, I have undone the merge as this seems to also contain a lot of other, unrelated changes.

Nedlinin:
@edde746 : Yea, thats on me! I accidentally included some extra code I was playing with in the commit and failed to check it before opening the PR.  I've opened a new PR with a minimal code change for this (https://github.com/edde746/plezy/pull/316)

PR #314, opened 2026-01-21
feat(video-controls): enable PiP support for macOS

svenjr:
Updates platform checks to include macOS alongside Android for PiP support, allowing the PiP button to appear and function on macOS. Helps address #44 

Screenshot 2026-01-20 at 10 31 52
Screenshot 2026-01-20 at 10 32 02

### DIFF
--- a/lib/widgets/video_controls/video_controls.dart
+++ b/lib/widgets/video_controls/video_controls.dart
@@ -619,7 +619,7 @@ class _PlexVideoControlsState extends State<PlexVideoControls> with WindowListen
 
   /// Check whether PiP is supported on this device
   Future<void> _checkPipSupport() async {
-    if (!Platform.isAndroid) {
+    if (!Platform.isAndroid && !Platform.isMacOS) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


edde746:
The issue with this approach is that it won't work on other fullscreen apps. The native macOS PiP window needs to be used.

svenjr:
> The native macOS PiP window needs to be used.

 I couldn't use the native macOS Picture-in-Picture API (AVPictureInPictureController) for several key technical reasons:

  1. Requires AVPlayerLayer: The native macOS PiP API only works with AVPlayerLayer (which is part of AVFoundation). This app uses MPV player (via MPVKit), which renders video using CAMetalLayer for hardware-accelerated playback.
  2. Can't Extract MPV's Video Layer: We discovered during implementation that MPV's video rendering is tightly coupled to its CAMetalLayer. When we tried to move this layer to a separate PiP window, the video rendering broke completely.
  3. Would Require Complete Rewrite: To use native PiP, you'd need to:
    - Replace MPV with AVPlayer entirely (losing all MPV features)
    - Or maintain two separate video playback systems (one for normal, one for PiP)
    - Both options would be massive architectural changes
  4. Android Uses Similar Approach: The existing Android implementation (PR #206) also doesn't use the native PictureInPictureParams in the traditional way - it adapts the approach to work with MPV.

edde746:
It's possible, see KSPlayer, IINA, etc. For now, closing this since it's severely limited.

PR #319, opened 2026-01-21
Show media release date

micahmo:
This PR adds the media release date to the metadata overlay. I checked the Plex Android, TV, and Desktop apps, and only the TV app shows the release date on the overlay, so this will elevate Plezy above the other Plex aps.

### Notes
- I add a new property, `originallyAvailableAt`, to the `PlexMetadata` model. Thankfully it was already provided in the API response. It returns the full release date of the media in the format `YYYY-MM-DD`, which I parse before displaying.
- Although both movies and TV shows have both a `year` and a `originallyAvailableAt` date, I chose to display the year only for movies and the full date for TV show episodes. This aligns with the Plex apps.
- The full date is shown using a region-sensitive format.

### Screenshots

|        | Movie | TV Show |
|--------|----------|----------|
| **Desktop** | image | image |
| **Mobile** | image | image |


edde746:
I feel like this might create some issues with the length of the string, you're already very close on the mobile TV show screenshot.

micahmo:
> I feel like this might create some issues with the length of the string, you're already very close on the mobile TV show screenshot.

That's a very good point. However, I did test and verify that the truncation via ellipses works fine with the longer string. And to be fair, even a long episode name alone could cause truncation on mobile.

**Episode Name Truncation**

image

**Date Truncation**

image

So maybe we are no worse off?

If you still don't like it, here are a few other ideas.
 - Put the date on a new line on mobile
 - Hide the date entirely on mobile
 - Show the date only for movies, not shows
 - Make the date optional on mobile (or just make it optional in general)
 - Or anything else? 😊

edde746:
Closing this, it's too much clutter to put on the video controls.

micahmo:
Totally understood, especially on mobile.

However, I think this is really valuable info, especially when we have more screen real estate (that's probably why Plex only puts it on their TV app).

Could we compromise? Here are a couple options that might make us both happy. 😊
 - Only show the date info when we're not on mobile? Then we have plenty of screen real estate and it doesn't feel cluttered. Plus the info is only visible when the overlay is showing, so it's not a constant distraction.
 - Or maybe we can just put the year for movies? Since the year is so short it's arguably no clutter at all.
 - Or we could keep this implementation as-is, but turn it off by default? Then it won't add any clutter for most users, but for those who want the info, they can go enable an option?

I'm willing to work to find a solution for this. 😊 Let me know what you think!

PR #321, opened 2026-01-22
Remember window state on desktop

micahmo:
This PR adds support for Plezy to remember its size, location, and maximized state between restarts of the app.

### Notes
 - I used the `window_manager` package that we already had to retrieve the window info. I added the `screen_retriever` package to help with validation (i.e., are the coordinates still valid for the current display(s)?).
 - The window events are only hooked up on when running on desktop.
 - There is a debounce timer on the window resized and window moved events since I figured those can fire quickly and we don't need to persist every pixel change.
 - If the user sets the size, then maximizes, then closes, then opens (maximized), then restores, it will have the original size (just a nice little QoL).
 - The `window_manager` package seems to struggle a bit with detecting coordinates on displays whose DPI/scaling is different than the other monitors.

### Demo

https://github.com/user-attachments/assets/5a67b17e-61c9-4264-bd60-777e1d67b278

### DIFF
--- a/lib/services/storage_service.dart
+++ b/lib/services/storage_service.dart
@@ -20,6 +20,11 @@ class StorageService extends BaseSharedPreferencesService {
   static const String _keyHiddenLibraries = 'hidden_libraries';
   static const String _keyServersList = 'servers_list';
   static const String _keyServerOrder = 'server_order';
+  static const String _keyWindowX = 'window_x';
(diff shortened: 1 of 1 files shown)


edde746:
macOS already handles this. I also think that this should probably be done purely natively to avoid the creation, then resize that can be seen in the video.

micahmo:
Closing in favor of #324.

PR #324, opened 2026-01-23
Remember window state on Windows

micahmo:
This PR adds support for Plezy to remember its size, location, and maximized state between restarts of the app on Windows using a native implementation. It supersedes #321.

### Notes
 - Only implemented for Windows.
 - No more flashing like the first implementation; now it opens immediately to the right place/size/state.
 - Works on my high-DPI monitor as well.
 - Still has validation to ensure we don't try to open on a missing monitor.
 - Still saves on all events with a debounce.
 - Implemented in such a way that we could add handling for other platforms (i.e., Linux) in the future.

### Demo Video

https://files.catbox.moe/5gde5c.mp4

(Video was too big for GitHub, sorry!)

### DIFF
--- a/lib/screens/main_screen.dart
+++ b/lib/screens/main_screen.dart
@@ -1,7 +1,9 @@
+import 'dart:async';
 import 'dart:io' show Platform, exit;
 
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:plezy/windows_native_window_state.dart';
 import 'package:provider/provider.dart';
 import 'package:url_launcher/url_launcher.dart';
 import 'package:window_manager/window_manager.dart';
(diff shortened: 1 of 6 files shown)


edde746:
Not really happy with how much code is used to achieve this, I feel like it could probably be done in an easier way. Are you using any AI tool to write this or just manually?

edde746:
Closed in favor of https://github.com/edde746/plezy/commit/bc9c2f53f2862cc6ec8da0205a11928645326f4a

micahmo:
> Closed in favor of [bc9c2f5](https://github.com/edde746/plezy/commit/bc9c2f53f2862cc6ec8da0205a11928645326f4a)

Thank you so much for implementing this! Your code is much more succinct. 😊

I just tested, and it works pretty well, but there is one small problem (that I also experienced when doing my implementation). It doesn't correctly maximize the window. Here's a demo. You can see it starts maximized and then goes down after that. I think it has to do with the timing of when the properties are applied.

https://github.com/user-attachments/assets/6e8fb93a-819d-4fed-95bb-43c2d33c44bc

The only way I could get this to work with my implementation was to call `::ShowWindow` with the `SW_SHOWMAXIMIZED` flag. I think it would be possible to add that to your implementation as well.

https://github.com/edde746/plezy/blob/ce50d261c49c5dad652093487ea6b03fb8782418/windows/runner/flutter_window.cpp#L54-L57

PR #326, opened 2026-01-24
Add episode release date to episode listing within season

micahmo:
This PR adds TV show episode release dates to the episode listing within the season view on desktop.

### Screenshots

| Plex | Plezy |
|--------|--------|
| image | image | 

### Notes
 - I added back the new `originallyAvailableAt` property to the `PlexMetadata` model from [#319](https://github.com/edde746/plezy/pull/319).
 - In my opinion, showing this is similar to showing the movie release year on the movie info page. This information is not available anywhere else in Plezy.
 - This change brings Plezy more in line with the Plex desktop app.

### Mobile
At this time I have **not** added the date on mobile yet because I couldn't get it to look good.

| Date inline | Date on new line |
|--------|--------|
| image | image | 

As you can see, when placed next to the duration and watch indicator, we quickly run out of room horizontally. I also tried placing it on its own line, but that make the episode metadata taller than the thumbnail, which doesn't look good.

If you have any other ideas, let me know, but for now I'm happy to have this on desktop.


edde746:
The watched indicator should probably be made better.

PR #366, opened 2026-02-02
Show unwatched episode count on shows/seasons

micahmo:
This PR adds a badge showing the number of unwatched episodes in a TV show for the Show and Season posters, in the same style as the watched indicator. This brings Plezy in line with Plex desktop, mobile, and TV apps.

### Plex Demo

https://github.com/user-attachments/assets/01ca705a-b5c3-4de0-a8a6-36e9b9394b84

### Plezy Demo

https://github.com/user-attachments/assets/b5fe257f-c7be-4e57-8fa3-fe12239f0711


edde746:
Would be nice to have a toggle for this in settings.

micahmo:
> Would be nice to have a toggle for this in settings.

Great idea! I added the option. It turned it on by default to match expectations for users coming from the Plex app. But if you think it should be off by default, I can do that instead.

image

PR #374, opened 2026-02-03
Collapse long show descriptions by default

micahmo:
This PR changes how long show descriptions are shown by collapsing them by default with the ability for the user to expand to read everything. I added this because I found that shows with long descriptions consume too much vertical real estate so that I have to scroll quite a ways before I can even see any of the seasons (see the "before" demo videos).

### Notes
- I allowed mobile to show more lines by default than desktop because the show/description is usually viewed vertically, so it would automatically consume more lines by default.
- On the desktop, the majority of shows (at least in my library) do not hit the threshold, so most people won't notice any difference in behavior here. On mobile (again, since everything is squished), it is more likely for regular shows to hit the threshold.
- I chose the `readmore` package because it's super easy to add text trimming by line count. It also allows handling clickable elements (i.e., links) which I'd like to add in a future PR.
- I'm not sure (and can't really test) how this will affect navigation on TV. Maybe I should exclude this for TV?

### Desktop - Before

https://github.com/user-attachments/assets/9d86a446-8e34-46bf-9991-e0781e2cae0c

### Desktop - After

https://github.com/user-attachments/assets/aca6fb69-230a-4274-a770-26a163b240b9

### Mobile - Before

https://github.com/user-attachments/assets/496b3b9d-412d-43f9-8bb3-aa30c8eb2603

### Mobile - After

https://github.com/user-attachments/assets/9bdab7d3-1add-465e-bfa4-5a391fdfe5ed

### DIFF
--- a/lib/screens/media_detail_screen.dart
+++ b/lib/screens/media_detail_screen.dart
@@ -1219,6 +1219,7 @@ class _MediaDetailScreenState extends State<MediaDetailScreen> with WatchStateAw
     final metadata = _fullMetadata ?? widget.metadata;
     final isShow = metadata.isShow;
     final isMobile = PlatformDetector.isMobile(context);
+    final isTv = PlatformDetector.isTV();
 
     KeyEventResult handleBack(FocusNode _, KeyEvent event) =>
         handleBackKeyNavigation(context, event, result: _watchStateChanged);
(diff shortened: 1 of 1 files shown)


micahmo:
> I'm not sure (and can't really test) how this will affect navigation on TV. Maybe I should exclude this for TV?

I was able to test this for TV and I did decide to disable this feature since it doesn't really make sense in that context.

PR #384, opened 2026-02-05
Add tooltip showing target time when hovering or dragging the playback slider

micahmo:
This PR adds a tooltip for when the user hovers over or drags the playback bar so that they can see what timestamp in the video that they will jump to.

### Notes
 - This is supported on desktop (when hovering with the mouse or dragging) and mobile (when dragging the slider).
 - I can see this mechanism also being expanded in the future to support [preview thumbnails](sliderPadding).
 - Please review with whitespace off for a more accurate representation of the changes.

### Demo

https://github.com/user-attachments/assets/d0fda702-f7ab-4c3a-9607-e50718b65607

https://github.com/user-attachments/assets/d58e70ae-f816-46f5-bb93-fd3a11ed0c05

### DIFF
--- a/lib/widgets/video_controls/widgets/timeline_slider.dart
+++ b/lib/widgets/video_controls/widgets/timeline_slider.dart
@@ -60,7 +60,7 @@ class _TimelineSliderState extends State<TimelineSlider> {
     final left = (pixelX - tooltipWidth / 2).clamp(0.0, sliderWidth - tooltipWidth);
     return Positioned(
       left: left,
-      top: -26,
+      top: -16,
       child: IgnorePointer(
         child: Container(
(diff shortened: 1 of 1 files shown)


edde746:
Slight nitpick, but I think it would look a lot better if it was further down/closer to the timeline.

micahmo:
> Slight nitpick, but I think it would look a lot better if it was further down/closer to the timeline.

Fair enough, I was experimenting with the positioning. Honestly the only reason I put it where I did is because of how big the playhead circle becomes when hovering. 

How does this look?

https://github.com/user-attachments/assets/a040c870-e4ce-406f-83fe-22d1a7b45f42

PR #448, opened 2026-02-11
Allow tv users to close control overlay and menus (subtitle, audio, etc.) with back button

dixon140:
This PR changes the back button functionality while in the player on Android TV. When the video has no control overlay, the back button exits the video. When the video has a control overlay, the back button will close the overlay. When in any of the menus (subtitle, audio, video, etc.), the back button will close the menu.

### DIFF
--- a/lib/screens/video_player_screen.dart
+++ b/lib/screens/video_player_screen.dart
@@ -163,6 +163,9 @@ class VideoPlayerScreenState extends State<VideoPlayerScreen> with WidgetsBindin
   final ValueNotifier<bool> _controlsVisible = ValueNotifier<bool>(
     true,
   ); // Track if video controls are visible (for popup positioning)
+  final ValueNotifier<bool> _isControlSheetOpen = ValueNotifier<bool>(
+    false,
+  ); // Track if a controls sheet is open (for TV back handling)
 
   @override
   void initState() {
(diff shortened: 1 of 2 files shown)


edde746:
Did you test this?

dixon140:
Yes I did, all features work as intended.

edde746:
https://github.com/user-attachments/assets/724c34d7-bff0-484d-a13e-e455f71ccbd0

Is this the intended behavior?

dixon140:
No, that's quite odd as that did not happen when I tested it on my fire stick. I will get back to you when I get home and review it.

dixon140:
@edde746 https://github.com/user-attachments/assets/e535cb6e-eb13-466c-bf56-007f7bb68483 I recloned my fork onto my laptop, and ran it wirelessly on my firestick. Seems to be functional. Maybe the emulator is causing issues? Not sure.

edde746:
Needs to work on all devices.

dixon140:
@edde746 Okay, after a lot of debugging, I think I fixed the issues. One issue seemed to be that the _controlsVisible variable would desync from the local player. I added code to fix this desync. The other issue was the way I was detecting if a menu was open, which I fixed by adding logic to keep track of their open/close state. I've tested on the android emulator, and it all seems to be working. I understand if it ended up being too many lines changed for such a small fix, but there was seemingly no consistent way to check if the controls were appearing, or if there was a menu present so I had to code extra to implement those. Thanks for reviewing!

PR #511, opened 2026-02-22
fix: airplane mode

Tazi0:
When you opened Plezy on airplane mode it would load for a while (trying to contact the server) before continuing to the offline_screen. This could be skipped if the system checks on if no wifi/data connection is found.

The addition code takes place in: `lib/main.dart` with `lib/screens/auth_screen.dart` for user feedback


edde746:
Quite a few unrelated changes, I also don't see the point in adding a new string.

Tazi0:
I added the pubspec changes because I had setup problems and updating fixed these.

Tazi0:
i cleaned up the translation

Issue #248, opened 2026-01-15
[Feature Request] Add an episode view


micahmo:
### Problem Statement

Right now, clicking an episode immediately jumps you into playback. On the official Plex web/mobile/desktop apps, there is a dedicated episode viewer page. It allows you to perform specific actions on the episode (e.g., mark as watched, etc.). I understand that these same actions can be performed via the context menu on the episode in the season view. But there's one other thing. Sometimes descriptions are long and you want to see/read the whole thing and maybe select text. Some descriptions have links and it's nice to be able to click the link as well.

### Proposed Solution

Add a dedicated episode view page.

### Alternatives Considered

Make the episode actions more obvious in the season view; allow the description to be interactable.

### Target Platform

All platforms

### Additional Context

I'd mainly like to be able to click on links in episode descriptions.

PR #538, opened 2026-02-25
Allow episode descriptions to be expanded

micahmo:
> Somewhat related to #248

This PR adds support for TV show episode descriptions to be expanded, instead of being hard-coded to 3 lines, using the same implementation as 6ac12e6ff9469fe9dde7bdc21fd065d6bad16240.

### Notes
 - For TVs, where it would be hard to focus the expander widget, the implementation is exactly the same as before (hard-coded cutoff to 3 lines).
 - I had to adjust the size of the expander pill (`...`) so that it would not consume too much vertical space when paired with `bodySmall`.
 - Most episode descriptions are already very short, so there should be very little noticeable difference. This is just to cover edge cases.
 - Please review with whitespace off.

### Demo

https://github.com/user-attachments/assets/c9b26a15-31d4-4fce-bb81-b15e9e349409

### DIFF
--- a/Casks/plezy.rb
+++ b/Casks/plezy.rb
@@ -1,6 +1,6 @@
 cask "plezy" do
-  version "1.21.3"
-  sha256 "81c643d9d67ed71ffa90d5ac36da6ef23fd0f71d3ae69f690cbe7710d511d74e"
+  version "1.22.0"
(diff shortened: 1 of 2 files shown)


edde746:
Screenshot 2026-02-27 at 22 19 49

This doesn't look aligned.

micahmo:
Let me know if this looks better.

| Before | After |
|--------|--------|
| image | image |

PR #561, opened 2026-02-26
Fix plex Auth notifications

8bitgentleman:
**Problem**

  Every time a user logged into Plex through Plezy, every server owner (including the user and any friends who share their server) received 3 back-to-back "New Device"
  notifications, and the device was identified as "Flutter" rather than "Plezy".

  **Root Cause**

  Two bugs, same underlying issue: connection probe requests were sent to Plex Media Servers without proper identification headers.

  When findBestWorkingConnection tests server reachability to pick the fastest endpoint, it runs multiple probes against each server across three phases: a cached
  endpoint test, an HTTPS upgrade attempt, and a latency sweep. All of these called testConnectionWithLatency, which created a bare Dio instance and only sent
  X-Plex-Token — no X-Plex-Client-Identifier, X-Plex-Product, or X-Plex-Device-Name.

  Plex Media Server registers a new device entry whenever it sees an authenticated request from an unrecognized X-Plex-Client-Identifier. Without that header, each of the
   3 probes appeared as a separate anonymous new device, triggering a notification for each one.

**Bug 1** 
 3 notifications per server: Each server gets probed 3 separate times during connection discovery with no client identifier, so the server owner receives 3 "New
  Device" notifications on every login — not just first install.

**Bug 2**
"Flutter" device name: With X-Plex-Product and X-Plex-Device-Name absent, Plex fell back to the raw HTTP user-agent (Dart/3.x (dart:io)) and displayed "Flutter"
   as the device name.

**Bug 3** 
auth_screen.dart was calling ServerConnectionOrchestrator.connectAndInitialize without a clientIdentifier. This caused MultiServerManager
   to fall back to DateTime.now().millisecondsSinceEpoch.toString() as the device ID — a brand-new throwaway string on every login. Even after fixing the probe headers,
  Plex would still register a fresh new device on every single login rather than recognizing the returning app.

  **How to Reproduce**

  1. Have a Plex account with at least one server (owned or shared)
  2. Log out of Plezy
  3. Log back in
  4. Check Plex notifications — 3 "New Device" notifications fire immediately, device shown as "Flutter (app name)"
  5. Log out and log in again — same 3 notifications fire again (not just on first install)

  **Fix**

  plex_client.dart — testConnectionWithLatency and testConnectionWithAverageLatency now accept an optional clientIdentifier and include X-Plex-Client-Identifier,
  X-Plex-Product: Plezy, and X-Plex-Device-Name: Plezy in every probe request.

  plex_auth_service.dart — findBestWorkingConnection (and the internal _upgradeCandidateToHttpsIfPossible) now accept and forward clientIdentifier through all probe call
  sites.

  multi_server_manager.dart — Both call sites of findBestWorkingConnection pass the stored clientIdentifier.

  auth_screen.dart — connectAndInitialize now receives _authService.clientIdentifier (the persistent UUID already loaded earlier in the same login flow), so probe
  requests during login use the same stable device ID as all other app sessions.

### DIFF
--- a/lib/screens/auth_screen.dart
+++ b/lib/screens/auth_screen.dart
@@ -97,6 +97,7 @@ class _AuthScreenState extends State<AuthScreen> {
         multiServerProvider: context.read<MultiServerProvider>(),
         librariesProvider: context.read<LibrariesProvider>(),
         syncService: context.read<OfflineWatchSyncService>(),
+        clientIdentifier: _authService.clientIdentifier,
       );
 
       if (!result.hasConnections) {


edde746:
I don't see the need of appName being a parameter.

8bitgentleman:
Sure, here's some more background. I have my own server and a friend has shared their server with me as well. Before this fix when I logged into plezy (as a dev this happens a bunch when testing new builds) both my friend and I would get 3 notifications from plex back to back. See the below screenshot for an example

Screenshot_20260225-212025

Now with this fix by assigning **appName**  it cuts the notifications down to one. When appName is not set the plex server sees each initial ping from plezy as coming from a different device. 

Screenshot_20260226-090758

PR #478, opened 2026-02-18
Download section rework - Auto download, retention settings, transcoding

Sphyix:
#241 and #213 

Added a few features
1- Automatic downloads of new episodes
2- Retention options for the downloads
3- Transcoding download with progress bar

Changes to the UI:

Download page:
Button to manually refresh entire series
Button to manually refresh single season
Button to access series download settings

Download settings:
Accessible when starting a new download and from the Settings button next to every series in the Download page
These settings take priority over App settings for automatic downloads.

App Settings page:
Switches to automatically download whole series and/or currently downloaded seasons.

----

Functionalities:
- Manual refresh buttons will follow download settings if set. If default, App Settings will be followed
- Saving new download settings will refresh the downloaded episodes.
- When changing quality, user is warned that transcoding to a new quality might take a long time.
- Starting a new download from the context menu or the series list will ask for settings.
- Last settings saved will be used as a default for new download settings when popup is opened.

Defaults:
- Previously existing downloads will go to default settings: Always keep all episodes, original quality
- Global settings defaults are: download all already existing seasons, do not add new seasons for a series.
Examples: 
- Season 2 of Series A is downloaded -> new episodes comes out -> default settings will download automatically
- Episode 1 of Season 2 of Series A is downloaded -> same happens
- Season 3 of Series A comes out -> default settings won't download that

----

Screenshots:

Downloads page with new buttons:
Screenshot 2026-02-18 at 03 06 14

Download settings:
image

Transcoding download with progress bar:
image

Quality change on downloaded episodes:
image

Download settings for movies:
Screenshot 2026-02-18 at 02 45 03

----

What's missing:
Playlist downloads - probably next PR

----

A good part of the work was done with AI, if it's not appreciated feel free to drop the PR.
I mostly did this since I have a 12+ hour flight soon, limited space on the SSD and a lot of anime to download :)

I tested most of this, hopefully I discovered all obvious bugs.
It's totally usable, I couldn't find many edge cases by myself.

PR #654, opened 2026-03-08
feat: auto-downloads, retention settings, download options

davidmckayv:
## Summary

Supersedes #478 (by @Sphyix) — includes all original changes plus review fixes.

**Features (from #478):**
- Per-series download settings: episode count, retention mode, transcode quality
- Auto-download new episodes on app resume with 5-min cooldown
- Smart episode queueing (keep last N unwatched)
- Watched episode cleanup with configurable retention (on refresh / after N days / after N weeks)
- Settings UI for global auto-download toggles
- Download settings dialog for shows and movies

**Review fixes applied:**
- AutoDownloadService singleton (lazy nullable + getter, matching codebase conventions)
- Transcode session matching by `TranscodeSession.key` instead of fragile `protocol == 'http'`
- Race condition guards: `.toList()` snapshot before async iteration, re-entrancy via `_queueing` set
- `findGlobalKeyForRatingKey` helper replacing fragile `endsWith(':$ratingKey')` patterns
- Move `_redownloadAtNewQuality` logic into `DownloadProvider`
- Split retention value into separate days/weeks state
- Orphaned settings cleanup on download deletion
- Loading spinner on refresh buttons with double-tap guard
- SharedPreferences cache for media version JSON blob deserialization
- Quality re-download snackbar feedback
- i18n keys added to all 15 locales
- Resolved merge conflicts with current main

## Test plan

- [x] Download a show with per-series settings, verify episode count and retention apply
- [x] Change quality on an existing download, confirm re-download prompt and execution
- [x] Switch retention between days/weeks, verify values are independent
- [x] Toggle auto-download settings, background the app, verify check runs on resume
- [x] Delete last episode of a show, verify orphaned settings are cleaned up
- [x] Tap refresh button rapidly, verify spinner shows and no double-requests
- [x] Test on desktop (macOS/Windows) — verify resumed lifecycle doesn't spam API calls

### DIFF
--- a/lib/providers/download_provider.dart
+++ b/lib/providers/download_provider.dart
@@ -308,9 +308,9 @@ class DownloadProvider extends ChangeNotifier {
           meta.grandparentRatingKey == showRatingKey &&
           progress != null &&
           (progress.status == DownloadStatus.completed ||
-           progress.status == DownloadStatus.downloading ||
-           progress.status == DownloadStatus.queued ||
-           progress.status == DownloadStatus.partial)) {
+              progress.status == DownloadStatus.downloading ||
+              progress.status == DownloadStatus.queued ||
+              progress.status == DownloadStatus.partial)) {
         final seasonRatingKey = meta.parentRatingKey;
         if (seasonRatingKey != null && !seasons.containsKey(seasonRatingKey)) {
           final seasonGlobalKey = '${meta.serverId}:$seasonRatingKey';
@@ -431,10 +431,7 @@ class DownloadProvider extends ChangeNotifier {
   }
 
   /// Get episode downloads filtered by show and/or season ratingKey.
-  List<DownloadProgress> _getEpisodeDownloads({
-    String? showRatingKey,
-    String? seasonRatingKey,
-  }) {
+  List<DownloadProgress> _getEpisodeDownloads({String? showRatingKey, String? seasonRatingKey}) {
     return _downloads.entries
         .where((entry) {
           final meta = _metadata[entry.key];
@@ -775,10 +772,7 @@ class DownloadProvider extends ChangeNotifier {
     notifyListeners();
 
     // Actually trigger download via DownloadManagerService
-    await _downloadManager.queueDownload(
-      metadata: metadataToStore,
-      client: client,
-    );
+    await _downloadManager.queueDownload(metadata: metadataToStore, client: client);
   }
 
   /// Fetch and store show and season metadata for an episode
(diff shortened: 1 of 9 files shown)


edde746:
The issue with transcoding downloads is that you need to keep the transcode session alive (or use the native, Plex Pass only download feature on the server). To keep the transcode session alive, you have to either re-write the download & transcode logic to be native on all platforms or drop the support for downloading in the background.

I haven't looked at it, but, do you download all audio & subtitle tracks? Because the normal Plex transcode will only send you 1 of each max.

davidmckayv:
Good points, both of these need to be addressed before this is ready.

**On keepalive:** Yeah, the current approach uses `protocol: http` which gives you a single continuous stream. The TCP connection itself acts as the keepalive, so there's no separate ping needed while actively downloading. But if the app backgrounds on mobile, the connection drops and the session dies in ~30s. You can't resume a dead transcode session either since a new one produces different bytes, so the partial file is wasted.

Even Plex's own app requires foreground for downloads ("File transfers cannot work if the app is moved to the background"), so foreground-only for transcoded downloads seems like the realistic near-term approach.

That said, I found that Plex now has a `/downloadQueue` API (server-local, documented at plexapi.dev) where the server handles transcoding in the background and you just poll until the file is ready, then download it as a normal file. That would work perfectly with `background_downloader` since the transcode and download are decoupled. Downside is it likely requires Plex Pass and a recent PMS version. Could be worth exploring as a future improvement though.

**On audio/subtitle tracks:** You're right, the universal transcoder only outputs 1 audio track and at most 1 subtitle track. There's no API workaround for that. The track that gets transcoded is whatever's currently "selected" on the server (controllable via `selectStreams` / PUT `/library/parts/{partId}`). Original quality downloads preserve all tracks obviously, but for transcoded downloads users would be stuck with a single audio/subtitle track. This should probably be documented clearly in the UI so users understand what they're getting.

What's your preference here? I'm thinking:
1. Transcoded downloads = foreground-only, with a clear warning about the track limitation
2. Original quality downloads = continue working in background as they do now
3. Maybe explore the `/downloadQueue` API later as a proper solution for both problems

edde746:
"there's no separate ping needed while actively downloading", incorrect. Stop relying on AI.

davidmckayv:
You're right, I was wrong about that. The server's 180s idle timer kills sessions regardless of HTTP activity, so periodic pings to `/video/:/transcode/universal/ping?session=<id>` (or timeline updates) are needed during the download. That's doable on a ~10-30s interval.

But that loops back to the same core problem you raised: you can ping while in the foreground, but once the app backgrounds on mobile the ping timer dies and the session gets killed. So the question still stands: foreground-only for transcoded downloads, or go with the `/downloadQueue` API to decouple transcoding from the download?

edde746:
image

Issue #668, opened 2026-03-11
MPV Errors, Black screen with audio


KenwoodFox:
### Bug Description

MPV seems to have some sort of hiccup where i see only black in the player, audio still plays. Wayland or X dosn't seem to change it and thats about all i tried besides using `vo=gpu` to get it to draw an extra box.

### Steps to Reproduce

Play any video
Observe the logs

### Expected Behavior

Video playback

### Actual Behavior

No video but audio is fine.

### Plezy Version

1.25.1

### Platform

Linux

### Device Information

Arch Linux

### Logs and Stack Traces

```
** Message: 00:05:22.148: MPV: Initialization successful (render context deferred)

** (com.edde746.plezy:53820): WARNING **: 00:05:22.149: MPV: No EGL context available
** Message: 00:05:22.391: MPV [fatal] vo/libmpv: No render context set.

** Message: 00:05:22.391: MPV [fatal] cplayer: Error opening/initializing the selected video_out (--vo) device.

[E]  [MPV:vo/libmpv] No render context set.
[E]  [MPV:cplayer] Error opening/initializing the selected video_out (--vo) device.
** Message: 00:05:22.601: MPV: Render context created with isolated EGL context
```

### Screenshots

<img width="1321" height="791" alt="Image" src="https://github.com/user-attachments/assets/7682e7b4-c0a0-4d26-943c-5a620e1cd752" />

edde746:
"Arch Linux", provide more information about your environment. How have you installed Plezy? Are you using X11 or Wayland?

KenwoodFox:
Ah im sorry, Im using wayland, kde plasma 6.6.2. Nvidia with the nvidia-open drivers, RTX 3080

```
❯ pacman -Qi plezy
Name            : plezy
Version         : 1.25.1-1
Description     : A beautiful Plex media server client built with Flutter
Architecture    : x86_64
URL             : https://plezy.app
Licenses        : GPL-3.0-or-later
Groups          : None
Provides        : None
Depends On      : alsa-lib  gtk3  libepoxy  libevdev  libkeybinder3  mpv
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 83.45 MiB
Packager        : Unknown Packager
Build Date      : Mon 09 Mar 2026 07:38:13 PM EDT
Install Date    : Tue 10 Mar 2026 11:22:38 PM EDT
Install Reason  : Explicitly installed
[...]

PR #669, opened 2026-03-11
Fix for #668

KenwoodFox:
This fixes the issue for me but might not be the right solution, after setVisible i 'think' mpv gets called too early and before drawing we get that 'no render context set' if we wrap a promise around scheduling just one frame it holds it and seems to work for me, since the context is ready by the time it continues.

Maybe that wait could be integrated into the setVisible? https://github.com/edde746/plezy/blob/4cc83006ecb512b64d5a2e57f5c7602c79de8c0f/lib/mpv/player/player_native.dart#L122

### DIFF
--- a/lib/mpv/player/player_native.dart
+++ b/lib/mpv/player/player_native.dart
@@ -1,5 +1,7 @@
+import 'dart:async';
 import 'dart:io' show Platform;
 
+import 'package:flutter/widgets.dart';
 import 'package:flutter/services.dart';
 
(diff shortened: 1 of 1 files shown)


edde746:
Wasn't the issue fixed by using the official binary?

KenwoodFox:
> Wasn't the issue fixed by using the official binary?

Kindof! It was fixed with bundling in the official binary but not when built standalone from `dart run...`

PR #683, opened 2026-03-12
Enhance Linux installation instructions in README

jianglai:
Added information about AUR packages for Linux installation.


edde746:
See #668

jianglai:
The plezy-bin package just downloads your .zst and repackages it:

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=plezy-bin

Seeing how frequent new releases are, having an AUR package makes the upgrade process much easier for users. Do you want to at least include a link to plezy-bin here?

PR #753, opened 2026-03-25
Fix missing season tabs

micahmo:
This is a small PR which fixes an issue where a single-season TV show would not display the season navigation tabs at the top. Note that this is for a library where Plex is configured to show seasons. (For libraries where Plex is not configured to show seasons, we correctly do not see the season tabs.)

### Library Settings

image

### Before

image

### After

image

### DIFF
--- a/lib/screens/media_detail_screen.dart
+++ b/lib/screens/media_detail_screen.dart
@@ -1204,7 +1204,7 @@ class _MediaDetailScreenState extends State<MediaDetailScreen>
       final isAlways = flattenSeasons == flattenSeasonsAlways;
       final isSingleSeason = flattenSeasons == flattenSeasonsSingleSeason;
       final shouldShowEpisodesDirectly =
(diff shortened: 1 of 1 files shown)


edde746:
The issue with this is that when the season tabs are show, they can also be focused, this adds an unnecessary "focus point" require one more input to get to the episodes.

micahmo:
Hmm, I see what you mean, but crucially it also removes the ability to perform actions on the season, like edit, delete, mark as watched, etc.

Is one extra focus point a big deal? We already have the ability to resume playback with the button at the very top, so I'd only need to arrow down if I wanted to examine a particular episode, and in that case, I'd most likely have to arrow through several other episodes before that. Plus an extra focus point is better than an extra click (which is what we had prior to the tabs).

Let me know your thoughts! Maybe we can figure out something else that works.

edde746:
You're right, needs to be focusable for those actions anyways.

PR #762, opened 2026-03-26
security: store X-Plex-Token in secure storage and strip from HTTP requests

rhinot:
## Summary

- Migrates `X-Plex-Token` and `X-Plex-Token` (Plex auth token) from `SharedPreferences` to `flutter_secure_storage` for encrypted at-rest storage on all platforms
- Adds a Dio interceptor that strips `X-Plex-Token` from outbound request headers, query parameters, and baked-into-path URLs before they leave the device — token is only added to requests targeting known trusted Plex server addresses
- Adds `PlexUrlHelper.isSecureDestination()` to validate that tokens are only sent to registered server URLs
- Extends the interceptor to cover the case where the token is embedded directly in `options.path` (e.g. via Plex channel URLs)
- Adds `resetForTesting()` / `testSecureStorage` injection point on `StorageService` for test isolation
- Adds `resetInstanceForTesting<T>()` to `BaseSharedPreferencesService` for generic singleton teardown in tests

## Tests

- `test/services/storage_service_test.dart` — save/retrieve/clear credentials including secure storage
- `test/services/plex_client_interceptor_test.dart` — security interceptor: header token stripped for untrusted hosts, query param token stripped, path-embedded token stripped, token preserved for trusted destinations
- `test/services/plex_auth_service_test.dart` — existing auth service tests (compile fix: added missing `dart:typed_data` import)

## Test plan

- [ ] Run `flutter test test/services/storage_service_test.dart`
- [ ] Run `flutter test test/services/plex_client_interceptor_test.dart`
- [ ] Run `flutter analyze` — no errors or warnings
- [ ] Verify token no longer appears in plaintext in app sandbox on device

### DIFF
--- a/lib/services/plex_client.dart
+++ b/lib/services/plex_client.dart
@@ -2235,7 +2235,7 @@ class PlexClient {
   }
 
   /// Get EPG channels using provider lineup endpoints (matches official Plex web client)
-  Future<List<LiveTvChannel>> getEpgChannels() async {
+  Future<List<LiveTvChannel>> getEpgChannels({String? lineup}) async {
     List<LiveTvChannel> parseChannels(Response response) {
       final container = _getMediaContainer(response);
       if (container != null && container['Channel'] is List && (container['Channel'] as List).isNotEmpty) {


rhinot:
Main PR that led to the other two PRs. I was checking that this app didn't send any information, whether personal or telemetry, other than what was needed for Plex to function. The result was cleaning information transmitted and securely storing tokens.

Two biggest changes
- Encryption at rest (currently plain text) in a standardized store
- Sensitive headers stripped so worst case information sent to a non-plex domain doesn't contain anything harmful

It's also supposed to handle migration of folks using current plain text creds.
Also added tests in case you decide to keep and improve.

Tradeoffs
- Slightly more memory usage
- Current implementation is probably too brittle (need to be able to start new db if old keys get corrupted)
- If plex decides to go to a different domain, it needs to be added

EPG change being included in this PR was an error.

Totally understand if you consider the tokens used not important enough to encrypt at rest. The header cleaning would be the main thing to consider, if just to prevent any errant sending of data.

rhinot:
I should also note that in auditing the code and network calls, nothing is sent, other than to plex. Thank you for creating a privacy conscious app!

PR #764, opened 2026-03-26
perf: batch metadata DB queries, add API cache getBatch, and misc cleanup

rhinot:
## Summary

- Adds `PlexApiCache.getBatch(Set<String> keys)` — single DB query replacing N+1 per-item lookups in `DownloadProvider`
- `DownloadProvider` now loads all pinned metadata in one pass on startup instead of querying per download
- Adds `AppDatabase.forTesting(super.connection)` and `PlexApiCache.forTesting(AppDatabase db)` named constructors for in-memory test isolation (no singleton side-effects in test runs)
- Caches `_downloadPathDisplay` in `SettingsScreen` state instead of re-fetching on every build; adds `mounted` guards to async callbacks
- Updates `.gitignore` to cover generated files, adds missing generated i18n strings in `strings.g.dart`
- Minor cleanup: `plex_metadata.dart` model update, `live_tv_screen.dart` / `guide_tab.dart` / `media_detail_screen.dart` / `episode_card.dart` / `hub_section.dart` / `media_card.dart` / `pin_entry_dialog.dart` tidy-ups

## Tests

- `test/services/plex_api_cache_test.dart` — 8 tests covering `getBatch`: empty set, no-match, single hit, multi-hit, partial hit, cross-server isolation, data round-trip

## Test plan

- [ ] Run `flutter test test/services/plex_api_cache_test.dart`
- [ ] Run `flutter analyze` — no errors or warnings
- [ ] Verify downloads list loads correctly on app startup
- [ ] Verify settings screen download path displays correctly

### DIFF
--- a/lib/database/app_database.dart
+++ b/lib/database/app_database.dart
@@ -16,6 +16,8 @@ part 'app_database.g.dart';
 class AppDatabase extends _$AppDatabase {
   AppDatabase() : super(_openConnection());
 
+  AppDatabase.forTesting(super.connection);
+
   @override
   int get schemaVersion => 8; // Added bgTaskId column to DownloadedMedia
 
--- a/lib/services/plex_api_cache.dart
+++ b/lib/services/plex_api_cache.dart
@@ -28,6 +28,9 @@ class PlexApiCache {
     _instance = PlexApiCache._(db);
   }
 
+  /// Create an isolated instance for testing (does not affect the singleton).
+  factory PlexApiCache.forTesting(AppDatabase db) => PlexApiCache._(db);
+
   /// Get the database instance (for services that need direct database access)
   AppDatabase get database => _db;
 


edde746:
Closed all your PRs since it seems they're all AI generated and you do not know what you're doing. You can re-open them if you wish, but you need to add a reason for each change, written in your own words.

rhinot:
Totally fair as the code was mostly Claude written & tested, where I guided and checked work.

This started as a security audit to ensure tokens or other information was not being leaked. Along the way, performance suggestions were made in the video player, then general other areas like database queries. I split the changes into 3 PRs (hence why I closed the single PR submitted before), so you can evaluate each. The tasks were added so you could see what I did to test. I also tested the PRs together in an emulator to look for things like hangs, stutter, etc.

This PR has the following changes:
- Removed overhead of spawning a new thread to write/query DB, as that is likely greater than the write of a view offset.  WAL (which you already had) prevents DB locks, so the thought process was this would cause less CPU load, therefore less stutter.
- New try logic reduces lookups from O(n) to O(1)
- More AI suggestions, which I understand if you want me to drop
   > FutureBuilder -> setState was admittedly a AI suggestion to get state without a rebuild
   > Android TV using native picker rather than just erroring out
- Lots of small cleanups:
   > Removed if else that had same output
   > Moved to _ DART convention as key was not necessary for function
   ...
   
I was trying to minimize number of PRs submitted, but happy to separate out just the DB changes, or you can just focus on the security changes and close these if you feel too much work to eval.

edde746:
The DB change was made to fix an ANR on Android.

PR #906, opened 2026-04-22
fix: register bundled GoNoto font in pubspec.yaml to fix NixOS rendering

eConnah:
Plezy seemed to have a bug where it relied on system fonts for rendering. As a result on nixOS without many fonts installed it resulted in invisible text. This PR changes the font declaration in pubspec.yaml according to this:https://docs.flutter.dev/cookbook/design/fonts and sets it as the default fontFamily in mono_theme.dart. I can confirm that this resulted in a build where I could see the font when using the nix derivation.

Before:
image

After:
image

However I am unsure why but in settings the header for each setting is still invisible, there are some fonts that are declared as fontFamily: monospace which is not packaged which may be part of the reason.

### DIFF
--- a/lib/theme/mono_theme.dart
+++ b/lib/theme/mono_theme.dart
@@ -43,6 +43,7 @@ ThemeData monoTheme({required bool dark, bool oled = false}) {
 
   final base = ThemeData(
     useMaterial3: true,
+    fontFamily: 'GoNoto',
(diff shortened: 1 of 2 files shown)


edde746:
GoNoto is bundled for subtitles, your PR removes it from the assets, likely breaking subtitles. This also sounds like an environment issue, not one with the app. Just install Roboto.

PR #910, opened 2026-04-23
Enhance auto-skip feature with animation control

swifty-tekno:
Added AnimationController for smooth auto-skip progress and updated related logic.

EDIT: should add more info
it removes the timer from the skip button text and replaces it by filling the skip button from left to right showing the progress

### DIFF
--- a/lib/widgets/video_controls/video_controls.dart
+++ b/lib/widgets/video_controls/video_controls.dart
@@ -283,7 +283,8 @@ class PlexVideoControls extends StatefulWidget {
   State<PlexVideoControls> createState() => _PlexVideoControlsState();
 }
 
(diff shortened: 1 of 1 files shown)


edde746:
The 200ms Timer.periodic it replaces was a deliberate frame-rate cap. The controls tree is big and sits over an active video player, so driving progress at refresh rate via setState on the State is a meaningful regression on TV hardware.

swifty-tekno:
oh okay i just made a changed to respect that cap but it looks like ass oh well il just leave it.
thanks for having a look anyway

EDIT:
looking into it i dont know how much it would affect but
adding a check to see if platform is tv and using old logic
and use the new logic on other platforms
would that be fine
im just looking for a way we could keep what i added while also not doing the regression on tv hardware like you said

PR #911, opened 2026-04-23
refactor for TV platforms

swifty-tekno:
i might have missed something but it should run like it did before on tv platforms with the new filling skip button on other platforms
i cant test on TV platform
i did flip the bools during testing to see if it would show old style for tv on desktop version and it showed fine like it did before

tbh i dont know what im doing im hoping i got everything so it respects the 200ms timer for tv

### DIFF
--- a/lib/widgets/video_controls/video_controls.dart
+++ b/lib/widgets/video_controls/video_controls.dart
@@ -580,22 +580,35 @@ class _PlexVideoControlsState extends State<PlexVideoControls>
 
     _autoSkipProgress = 0.0;
 
-    // Smooth progress for the Netflix-style wipe: drive progress every frame.
-    // We still keep marker identity checks so the animation never leaks across markers.
     _autoSkipController?.dispose();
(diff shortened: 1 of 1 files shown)


edde746:
- The new debounce in `_performAutoSkip` reuses `_lastSkipActionTime`, which is already owned by the seek double-tap handlers — these unrelated flows will now cancel each other within 200ms.
- Non-TV path does per-frame `setState` via `addListener`, violating the same frame-budget rationale the TV branch preserves. `AnimatedBuilder` would avoid the whole-subtree rebuild.
- Redundant `ignoring: true`, redundant null check, `PlatformDetector.isTV()` called 4× per build, title doesn't match where the real change happened.

swifty-tekno:
is this better as i said before i dont really know too much i was just trying to add a nice thing to the skip button and forgot about other platforms

https://github.com/swifty-tekno/plezy/commit/3ad8678d8a6d7ee218319e0639c2989502a4a621

edde746:
`_toggleControls` no longer cancels auto-skip — on touch devices this means tap-to-see-controls doesn't interrupt the countdown anymore.

PR #926, opened 2026-04-26
Show season posters on tab in show view

micahmo:
I really like the tab-style navigation for seasons and how you get a flat list of episodes by default, but I was lamenting that I don't really see season posters anymore, so I was trying to find a way to integrate them without taking up too much vertical real estate. Notably, this change does not affect navigation or anything like that. If this is to much and/or you don't like it, no worries, at all. Just an idea. 😊

https://github.com/user-attachments/assets/70e0f49b-6d1d-4b33-8c30-d591f472f273


edde746:
Can merge if you gate it behind a setting (off by default).

micahmo:
Done!

https://github.com/user-attachments/assets/c7999587-245a-4a0c-bef2-e8442b0b8c47

PR #928, opened 2026-04-27
Improve Android back button behavior

micahmo:
This PR makes two changes related to Android back button handling.

1. When any tab other than "Home" is selected, pressing back will navigate to home.
2. When on the home screen, pressing back will show a snackbar, indicating that you must press back once more to exit (this is a common pattern on Android to help prevent accidental closing of the app).

> Review with whitespace off.

https://github.com/user-attachments/assets/c2e54110-6c76-420d-8a8e-da4c8a03b7a0

### DIFF
--- a/lib/i18n/da.i18n.json
+++ b/lib/i18n/da.i18n.json
@@ -55,7 +55,8 @@
     "startingOfflineMode": "Starter offlinetilstand...",
     "loading": "Indlæser...",
     "fullscreen": "Fuldskærm",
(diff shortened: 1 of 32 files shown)


crakerjac:
I like this.  I cant tell you how many times I've hit the back button to go back a screen only to end up on the home screen.

edde746:
Implemented this in another commit with better handling for TV, forgot to close this.

PR #949, opened 2026-04-29
fix(linux): persistent HTTP/1.1 connection pool to prevent parallel-request timeout

### LINKED ISSUE #929: 'PlexHttpException(connectionError: No route to host)' when Secure Connections set to Required in Plex settings
I posted a comment in a related issue that you fixed, but the new issue is a bit different.

> @edde746 Thanks for the quick fix - the `type 'String' is not a subtype` issue is gone. However, the underlying home screen stall still occurs, just for a different reason.
> 
> When a timeout triggers failover (with Plex still set to secure connections only), the app cycles through local/plex.direct candidates - which all return `connectionError: No route to host` - before eventually landing back on the working HTTPS custom URL. During that cycle the home screen stops loading.
> 
> 
> <details>
> <summary>Relevant log snippet (1.35.0, Linux/Arch on WiFi)</summary>
> 
> ```
> [19:07:33] Switching Plex endpoint after GET failure
>           {from: https://[custom-domain]:443, to: https://192-x-x-170.[plex.direct]:32400, path: /hubs/sections/6}
> [19:07:39] Failed to get library hubs: PlexHttpException(connectionError: No route to host)   <- plex.direct:32400 unreachable
> [19:07:39] Switching Plex endpoint: {from: https://192-x-x-170.[plex.direct]:32400, to: http://192.x.x.170:32400}
> [19:07:42–19:07:57] Failed fetching on deck / library hubs (×4 cycles): PlexHttpException(connectionError: No route to host)   <- local IP unreachable
> [19:07:57] All endpoints exhausted for [server], triggering reconnection
> [19:08:00] Successfully reconnected via https://[custom-domain]:443   <- back to the working URL
> [19:08:01] Fetched 20 on deck items ✓   <- home screen finally loads
> ```
> 
> </details>
> 
> **Setup:** Plex behind Caddy reverse proxy, custom URL `https://[domain]:443`, Secure Connections = Required. Notably this doesn't happen on Android on the same WiFi network, so it may be Linux-specific.
> 
> It looks like the failover priority/filtering doesn't account for whether local/plex.direct endpoints are actually reachable from the current network. Would it help or is it possible to skip candidates that are IP-local or on non-standard ports when the preferred endpoint is an external HTTPS hostname? 

 _Originally posted by @bl4ckswordsman in [#900](https://github.com/edde746/plezy/issues/900#issuecomment-4329399903)_

bl4ckswordsman:
Problem: Linux lacks a native HTTP/2 client in the Dart ecosystem ([dart-lang/http#1385](https://github.com/dart-lang/http/issues/1385)). Every PlexHttpClient created a fresh IOClient, meaning each of the 10+ parallel home-screen requests (on-deck, hubs, metadata) opened a separate TCP+TLS handshake. Dart's default maxConnectionsPerHost = 6 caused trailing requests to queue past the 10s connect timeout, triggering spurious failover to unreachable LAN/plex.direct candidates.

Additionally, PlexHttpClient.close() called IOClient.close() which called HttpClient.close() internally - destroying all keep-alive connections and forcing cold handshakes on every reconnect.

Fix:

- Added a file-level HttpClient singleton for Linux with maxConnectionsPerHost = 12 and idleTimeout = 90s. createPlatformClient() wraps it so all PlexHttpClient instances share one pool.

- Added closePlexClient() - no-op on Linux, normal close() elsewhere. PlexHttpClient.close() now delegates to this instead of calling _client.close() directly.

- ~~Added createProbeClient() - returns a fresh, disposable IOClient for connection testing that cannot accidentally close the singleton pool.~~

- Stub versions of all three functions added to platform_http_client_stub.dart for web compile compatibility.

Android/iOS/Windows behaviour is unchanged.

Also fixes #929 in my testing.

### DIFF
--- a/lib/services/plex_client.dart
+++ b/lib/services/plex_client.dart
@@ -214,13 +214,9 @@ class PlexClient with MediaServerCacheMixin implements MediaServerClient {
       client._serverTranscoderCached = seedTranscoderVideoSupport;
     }
     await client._initMediaProviders();
-    // If the connection race didn't seed the capability, warm the cache in
-    // the background so the first playback doesn't pay the probe cost on its
-    // hot path.
-    if (seedTranscoderVideoSupport == null) {
-      unawaited(client.serverSupportsVideoTranscoding());
-    }
-    // Don't fire background work if the caller already timed out
+    // If the connection race didn't seed the capability, or if the caller
+    // already timed out, warm the cache in the background so the first playback
+    // doesn't pay the probe cost on its hot path
     if (isCancelled?.call() != true && seedTranscoderVideoSupport == null) {
       unawaited(client.serverSupportsVideoTranscoding());
     }
--- a/lib/services/plex_client.dart
+++ b/lib/services/plex_client.dart
@@ -173,6 +173,7 @@ class PlexClient {
     Future<void> Function(String newBaseUrl)? onEndpointChanged,
     VoidCallback? onAllEndpointsExhausted,
     bool? seedTranscoderVideoSupport,
+    bool Function()? isCancelled,
   }) async {
     final client = PlexClient._(
       config,
@@ -190,6 +191,10 @@ class PlexClient {
     // the background so the first playback doesn't pay the probe cost on its
     // hot path.
     if (seedTranscoderVideoSupport == null) {
+      unawaited(client.serverSupportsVideoTranscoding());
+    }
+    // Don't fire background work if the caller already timed out
+    if (isCancelled?.call() != true && seedTranscoderVideoSupport == null) {
       unawaited(client.serverSupportsVideoTranscoding());
     }
     return client;
@@ -308,7 +313,7 @@ class PlexClient {
   /// This discovers individually shared items that don't appear in /library/sections.
   Future<void> _initMediaProviders() async {
     try {
-      final response = await _getWithFailover('/media/providers');
+      final response = await _getWithFailover('/media/providers'); //.timeout(const Duration(seconds: 6));
       final container = _getMediaContainer(response);
       if (container == null) {
         _providerLibraries = [];
--- a/lib/services/plex_client.dart
+++ b/lib/services/plex_client.dart
@@ -200,6 +200,7 @@ class PlexClient with MediaServerCacheMixin implements MediaServerClient {
     Future<void> Function(String newBaseUrl)? onEndpointChanged,
     VoidCallback? onAllEndpointsExhausted,
     bool? seedTranscoderVideoSupport,
+    bool Function()? isCancelled,
   }) async {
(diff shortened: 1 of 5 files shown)


bl4ckswordsman:
Hey  @edde746 could I please get some input, is this not up to standards?

It's only 35 lines, linux-specific, and fixes some timeouts related to caddy/http2 use cases which I constantly get without this patch.

edde746:
Been busy with adding Jellyfin support, I believe there was an issue with doing this which led to me switching to native HTTP clients where available. Also:

1. `createPlatformClient()` now returns an `IOClient` backed by the Linux singleton, but several non-`MediaServerHttpClient` callers still call `.close()` directly. For example, Fribb refresh/download and Trakt disposal can close the shared `HttpClient`, leaving `_linuxSingleton` pointing at a closed client. Later requests would then fail unexpectedly.
2. The transcoder warm-up cancellation guard currently does not work as intended because `serverSupportsVideoTranscoding()` is still launched once before the guarded block. It also means the probe can be launched twice when `seedTranscoderVideoSupport == null`.
3. The PR mentions `createProbeClient()`, but the probe paths do not appear to use it yet. Either those paths should be wired to a disposable client, or the helper should be removed from this PR.

bl4ckswordsman:
My bad for disturbing and interrupting your flow. And thanks for the input.

I removed the duplicate probe fire and unused `createProbeClient()`.

Regarding the `createPlatformClient` references.
`fribb_mapping_store.dart` creates short-lived clients in two download methods and almost certainly closes them in a `finally` block, which would destroy the singleton on Linux. The tracker service constructors (`TraktClient`, `MalClient`, `SimklClient`, `AnilistClient`, `OAuthProxyClient`, `DeviceCodeAuthServiceBase`) store the result as a field and would do the same on disposal.

Since these callers have seemingly no reason to share the Plex connection pool,  perhaps they should use a plain disposable client instead(?):

```dart
// fribb_mapping_store.dart and all tracker service constructors
- _http = httpClient ?? platform.createPlatformClient();
+ _http = httpClient ?? http.Client();
```

`http.Client()` on Linux would then resolve to a plain `IOClient` with its own `HttpClient`, so behaviour is unchanged, it just can't accidentally close the singleton.

What do you think about this? (whenever you have a few minutes to spare ofc)

I wanted to keep the patch clean and the scope of this PR as narrow as possible. But if it gets too broad or too complicated, and since you mentioned you had issues with this approach before, I might just have to give up on it and keep building a patched version on my own.

Thanks anyways!

edde746:
Closing in favor of https://github.com/edde746/plezy/commit/fb084d3ea1f349b4f8527404ecd16b81ce554fbf

Issue #1075, opened 2026-05-18
2.1.1 - Playback breaks after pausing or entering PiP on iOS


jmswangit:
### Bug Description

I noticed this on 2.0.0 but waited to report because I figured the next update would fix it, but it’s been persistent. When I play an item in my plex library, playback works great - until I pause the media or enter Picture in Picture mode. When I do either of those and play, the picture freezes. The screen unfreezes if I remain in the app and I rewind or go forward, but this workaround does not work on PiP mode. This issue occurs on every media item I’ve tested. 

### Steps to Reproduce

Play an item on iOS
Pause the item
Press play
Notice screen has frozen. 

Alternatively:
Play an item on iOS
Enter PiP mode
Notice screen has frozen.

### Expected Behavior

Not freezing

### Actual Behavior

Freezing

### Plezy Version

2.1.1

### Platform

iOS

### Media Server

Plex

### Device Information

iPhone 16 Plus

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

edde746:
Which iOS version?

jmswangit:
I am using an iPhone 16 Plus on iOS 26.4.

edde746:
Could you enable Debug Logging and reproduce:
1. Start playback
2. Pause
3. Wait a few seconds
4. Play
5. Enter PiP if possible

Then upload/copy logs from Settings -> View Logs and share the log ID?

jmswangit:
Log ID: o4m3y

razzor269:
Same issue for me since version 2.0.0. I noticed a lot of the following error in the debug log after reproducing the problem:

[INFO] [MPV:vo/avfoundation] Display layer not ready after bounded wait; dropping frame

Only happens when I am pausing the playback and then try to continue.

edde746:
https://testflight.apple.com/join/AUUPnQDh, try this build

jmswangit:
Just got it downloaded and the pip issue is completely gone and the pause issue has reduced significantly to just a very very minor stutter. This is good enough for me, but I just wanted to let you know. Log ID: 2hx0b

razzor269:
I can confirm that the problem is almost completely gone with your beta version. After pressing play the picture stood still for a brief moment and then the playback continues without any further issues.

edde746:
Pushed a new build that should fix the stuttering

jmswangit:
Confirmed! The issue is completely gone now. Thanks a million!

PR #1111, opened 2026-05-23
feat(player): add end-of-video sleep timer option

### LINKED ISSUE #540: Add “End of current video” option to Sleep Timer (like Plex)
### Problem Statement

The current sleep timer works based on fixed time intervals (e.g., 5, 15, 30 or 60 minutes).

While this works well, there is no option to stop playback after the current video finishes. Adding this would improve the overall sleep timer experience.

### Proposed Solution

Add a new sleep timer option:
“End of current video”

When enabled:
- Let the current video finish completely
- Stop playback before the next episode starts

This works similarly to how Plex handles sleep behavior.

### Alternatives Considered

_No response_

### Use Case

1. As a user, I would enable “End of current video” before sleeping.
2. This would let the episode or movie finish naturally.
3. The benefit would be no mid-scene interruptions and a smoother sleep experience.

### Target Platform

Android, All platforms

### Additional Context

This would improve the overall sleep timer experience, especially for users watching long movies or episodic content with autoplay enabled.

Image

timokox:
## Summary

Adds an **"End of current video"** entry to the sleep timer list, sitting above the existing fixed-duration options (5/10/15/30/45/60/90/120 min). When chosen, playback pauses at the end of the currently playing item — overriding autoplay and the next-episode dialog.

Closes #540.

## Why

The existing minute-based timer forces a choice between cutting off mid-scene (timer too short) or continuing into the next episode and then cutting off mid-scene anyway (timer too long). The previously suggested workaround — disabling autoplay globally — has to be toggled back and forth depending on whether the user wants to binge or wind down, and still doesn't actually stop the player at end of file. This option solves it cleanly without changing any existing defaults.

## Design

End-of-video mode is a separate path in `SleepTimerService` rather than a shoehorned zero-duration timer:

- No periodic `Timer` runs. Playback completion is reported externally via `notifyVideoCompleted()`, called from `_onVideoCompleted` in `playback_prompts.dart`.
- The check sits at the top of `_onVideoCompleted` so it pre-empts the next-episode dialog and the autoplay countdown — the user explicitly asked to stop after this item.
- The mode is preserved across in-place player teardown (`markNeedsRestart` / `restartIfNeeded`), mirroring the duration-timer behavior, so quitting and starting a new item keeps the timer armed.
- `isActive` covers both modes; `isEndOfVideoMode` is exposed so the UI can render the right active-status variant.

## UI

- **`SleepTimerDurationList`** — new entry with a `hourglass_bottom_rounded` icon at index 0, above the minute options.
- **`SleepTimerActiveStatus`** — renders both modes from one widget. Duration mode keeps the countdown text and the "+15 min" extend button. End-of-video mode shows a static "Playback will pause at the end of this video" message with only the cancel button (extending a non-time-based timer makes no sense).
- **`SleepTimerContent`** — show-active-status guard now also triggers when `isEndOfVideoMode` is true (since `remainingTime` is null in that mode).
- Snackbar confirmation reuses `t.messages.sleepTimerSet` with the localized label.

## i18n

Added two new strings to `lib/i18n/en.i18n.json` and `lib/i18n/de.i18n.json`:

- `videoControls.sleepTimerEndOfVideo` — "End of current video" / "Ende des aktuellen Videos"
- `videoControls.playbackWillPauseAtEnd` — "Playback will pause at the end of this video" / "Wiedergabe wird am Ende dieses Videos pausiert"

Other locales (da, es, fr, it, ja, ko, nb, nl, pl, pt, ru, sv, zh) fall back to en via `fallback_strategy: base_locale` already set in `slang.yaml` — no missing-implementation errors from the generator.

## Tests

Eight new unit tests in `test/services/sleep_timer_service_test.dart` covering:

- `armEndOfVideo` bookkeeping (sets isActive + isEndOfVideoMode, leaves duration/endTime null)
- replacement of a running duration timer
- cancel clears mode
- listener notifications
- `notifyVideoCompleted` fires the callback + emits `onCompleted` + consumes the mode
- no-op when not armed (also doesn't interfere with a running duration timer)
- `restartIfNeeded` preserves the end-of-video mode across a session swap

Full suite (1792 tests) green locally. `dart format .`, `flutter analyze` (no new issues), `flutter test` all clean.

## Testing notes

- Tested locally on iOS via `flutter test`. Manual smoke test on a physical device is recommended before release — particularly the interaction with the existing "Play Next" dialog at the end of an episode that has a next episode queued.
- The end-of-video trigger relies on `currentPlayer.streams.completed` (already used by autoplay), so it inherits the same EOF detection.
- The fix in #1075 (iOS PiP/pause freeze) does not affect this path.


edde746:
> sitting above the existing fixed-duration options

I'd prefer if it was implemented with a split panel like the subtitle/audio sheet.

timokox:
Reworked the sheet to use the split-panel layout from `TrackSheet` — two columns inside a `Row`, each in its own `FocusTraversalGroup`:

- **Left ('Stop at')** — event-based stop options. Today just 'End of current video'; left open-ended for future triggers (end of next episode, end of chapter, …).
- **Right ('Timer')** — the existing fixed-duration options, unchanged.

Active option is highlighted inline via `FocusableListTile.selected` in whichever column it belongs to. Pushed as a separate commit (`refactor(player): sleep timer sheet uses split-panel layout`) for easier review; happy to squash before merge if you prefer.

PR #1129, opened 2026-05-25
feat(jellyfin): add Unwatched filter when browsing libraries

RyoShinzo:
## Summary

Jellyfin's library filter panel currently exposes only Genre, Year, Content Rating, and Tag — no way to filter to unwatched movies/shows. Plex already exposes this as a boolean toggle. This PR brings parity by adding an "Unwatched" toggle for Jellyfin libraries.

## What was discovered

`JellyfinClient.fetchLibraryFiltersWithValues` (`lib/services/jellyfin_client/parts/browse.dart`) builds its filter list purely from Jellyfin's `/Items/Filters` response, which has no boolean-toggle equivalent. But the surrounding plumbing was already in place:

- `JellyfinLibraryQueryTranslator.toQueryParameters` already maps `includeWatched: false` → `Filters=IsUnplayed`.
- `filters_bottom_sheet.dart` already renders `filterType == 'boolean'` entries as toggle switches.
- `PlexLibraryQueryTranslator` already uses the key `unwatched`.

So the filter could be added with a small hardcoded entry — no translator changes, no UI changes.

## How it was resolved

`fetchLibraryFiltersWithValues` injects a hardcoded `MediaFilter(key: 'unwatched', filterType: 'boolean', ...)` into its returned list, modeled on how `fetchSortOptions` hardcodes its options. The key matches the existing Plex key so the shared translator picks it up.

To keep the toggle off music/photo libraries where `IsUnplayed` is meaningless, a `libraryType` parameter was threaded through `MediaServerClient.fetchLibraryFiltersWithValues` (mirroring the existing `fetchSortOptions` pattern). The Jellyfin implementation only emits the Unwatched filter when `libraryType == 'movie' || libraryType == 'show'`. Both call sites (`LibraryFilterSortLoader`, `LibraryBrowseTab._loadJellyfinFiltersInBackground`) pass `library.kind.id`. The Plex implementation accepts the param and ignores it — Plex's filter list is server-driven.

## What to look out for in review

- **Interface change**: `MediaServerClient.fetchLibraryFiltersWithValues` gained an optional `String? libraryType`. Both backends and both call sites are updated; no other callers exist.
- **i18n**: adds one new string (`en.i18n.json`) for the toggle label; regenerated `strings*.g.dart` is included.
- **Filter key coupling**: the literal `'unwatched'` is shared between the Plex translator (existing) and the new Jellyfin entry. Renaming it would require updating both sides.
- **Verification**: code-review + `dart format` + `flutter analyze` clean on touched files + full `flutter test` (1795 tests) green. **Manually verified on Linux desktop only** (toggle appears on movie/show libraries, hidden on others, filters correctly). Contributor doesn't have access to iOS/tvOS for visual retest — the code paths are platform-independent (no platform branching in the diff) so behaviour should be uniform, but flagging in case the maintainer wants to spot-check.

Background: [RyoShinzo/plezy#1](https://github.com/RyoShinzo/plezy/issues/1)

### DIFF
--- a/lib/i18n/en.i18n.json
+++ b/lib/i18n/en.i18n.json
@@ -705,7 +705,8 @@
       "genre": "Genre",
       "year": "Year",
       "contentRating": "Content Rating",
-      "tag": "Tag"
+      "tag": "Tag",
+      "unwatched": "Unwatched"
     },
     "sortLabels": {
       "title": "Title",
--- a/lib/i18n/strings.g.dart
+++ b/lib/i18n/strings.g.dart
@@ -4,7 +4,7 @@
 /// To regenerate, run: `dart run slang`
 ///
 /// Locales: 15
-/// Strings: 17565 (1171 per locale)
+/// Strings: 17566 (1171 per locale)
 
 // coverage:ignore-file
 // ignore_for_file: type=lint, unused_import
--- a/lib/i18n/strings_en.g.dart
+++ b/lib/i18n/strings_en.g.dart
@@ -3775,6 +3775,9 @@ class TranslationsLibrariesFilterCategoriesEn {
 
 	/// en: 'Tag'
 	String get tag => 'Tag';
+
+	/// en: 'Unwatched'
+	String get unwatched => 'Unwatched';
 }
 
 // Path: libraries.sortLabels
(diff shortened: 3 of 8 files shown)


edde746:
Closed in favor of https://github.com/edde746/plezy/commit/366ae67341a56b436e569b590305b38f4316f7b6

PR #1147, opened 2026-05-26
feat(tracker): implement watch state authority for trackers

afalfallaj:
## Tracker Watch State Authority

Adds support for using a connected tracker as the authoritative source for watch history — watched status, progress, Continue Watching, and the Playlists tab — replacing Plex's native data before it reaches the UI. Trakt is the first tracker to support this; the architecture is built to accommodate others.

---

### Motivation

Plex only knows what you watched through Plex. If you watch something on another app, device, or client, that progress does not exist in Plex. A tracker that aggregates across all clients does. This feature surfaces that data in Plezy without requiring the user to reconcile two watch histories manually.

---

### Features

- **Unified watch history** — watched badges and resume points reflect everything you've watched, not just what Plex saw
- **Cross-client Continue Watching** — pick up where you left off even if you started an episode on another device or app
- **Tracker watchlist as Playlists** — your curated tracker list takes over the Playlists tab, putting what you actually want to watch front and center
- **One-tap activation** — a single toggle in the tracker's settings turns the feature on or off
- **Non-destructive** — flipping the toggle off instantly restores Plex-native data; nothing is migrated or rewritten

---

### What is not affected

**Plex watch data is read-only from Plezy's perspective.** The overlay replaces data at render time; nothing writes to or modifies Plex's own watch state. The tracker's watchlist is likewise read-only — Plezy never mutates the list on the tracker side.

All other features — library browsing, metadata, playback, search — are unaffected. The overlay is applied as a final step before the UI renders and has no side effects beyond watched fields and the Playlists tab content.

Users who do not enable the toggle see no behavioral change.

---

### How it works

A small shared layer sits between Plex and the UI. When a tracker is connected and turned on, that layer quietly swaps Plex's watch data for the tracker's right before the screen renders — that's it.

The UI never talks to a specific tracker. It just asks the shared layer for the latest data. That means adding another tracker later (Simkl, AniList, MAL) is a small job, not a rewrite — fill in one contract and it works everywhere.

---

### Testing

AI assisted in building this feature, but every step was reviewed and tested in code by me. This has been running in daily use for several weeks across multiple devices and libraries. The following paths were verified:

- Cold boot with authority enabled and a large tracker history
- Profile switching mid-session
- Offline playback — tracker data is served from cache; stub actions (e.g. Remove from Continue Watching) are blocked with a clear message rather than silently failing
- Toggling authority on and off at runtime, with screens reverting to Plex-native data immediately
- Continue Watching items started outside Plezy resolving to the correct Plex episode
- Playlists tab swaps to the tracker's watchlist when authority is on and back to Plex playlists when off; items resolve only to entries the user actually has in their Plex library

---

### What's next

Trakt ships first. Simkl, AniList, and MAL can plug into the same layer when we're ready — the groundwork is already done.

---

Let me if you have feedback on any part of this scope, architecture, or specific pieces that don't fit the direction you have in mind for the project.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -36,9 +36,8 @@ import 'services/discord_rpc_service.dart';
 import 'package:path_provider/path_provider.dart';
 import 'services/image_cache_service.dart';
 import 'services/gamepad_service.dart';
-import 'services/trakt/trakt_scrobble_service.dart';
-import 'services/trakt/trakt_sync_service.dart';
-import 'services/trackers/tracker_coordinator.dart';
+import 'services/trackers/tracker_lifecycle.dart';
+import 'services/trackers/watch_state_overlay.dart';
 import 'providers/trakt_account_provider.dart';
 import 'providers/trackers_provider.dart';
 import 'providers/user_profile_provider.dart';
@@ -245,7 +244,7 @@ Future<void> _bootstrapApp() async {
     unawaited(DiscordRPCService.instance.initialize());
   }
 
-  await TraktScrobbleService.instance.initialize();
+  await TrackerLifecycle.initializeScrobble();
 
   _registerShaderLicenses();
 
@@ -483,9 +482,7 @@ class _MainAppState extends State<MainApp> with WidgetsBindingObserver {
 
     _offlineWatchSyncService = OfflineWatchSyncService(database: _appDatabase, serverManager: _serverManager);
 
-    // Trakt sync service (subscribes to WatchStateNotifier, requires serverManager
-    // to resolve PlexClients for GUID lookups).
-    TraktSyncService.instance.initialize(serverManager: _serverManager);
+    TrackerLifecycle.initializeWithServerManager(_serverManager);
 
     _appLifecycleListener = AppLifecycleListener(
       onExitRequested: () async {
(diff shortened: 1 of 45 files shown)


edde746:
Closing this PR.

The feature direction does not fit Plezy in its current form. This moves too much watch-state authority into the client, adds broad tracker/Plex reconciliation logic, relies on brittle Trakt behavior, and creates a long-term maintenance surface around cache state, profile switching, offline behavior, multi-server matching, anime numbering, and action semantics.

The current diff is also too large to review safely. Several correctness risks showed up during review, including cache failure handling, profile-switch races, library-filter bypasses for progress pushes, unresolved offline stubs, and “Remove from Continue Watching” changing watch history.

The apparent support activity around this PR also raises trust concerns. A newly involved account posting support and reacting on the PR, while appearing closely tied to the author, reads like coordinated or undisclosed support rather than independent feedback. That is not acceptable for influencing project direction.

Please also clarify whether AI assistance was used to generate or substantially implement this PR. The implementation appears heavily AI-assisted, and that should be disclosed plainly, especially for a change of this size and maintenance impact.

This would need a design/RFC first, with much smaller scoped changes, clear product boundaries, no surprising replacement of existing Plex features, transparent disclosure around authorship/support, and targeted tests before being reconsidered.

afalfallaj:
Thanks for taking the time to review this. I see where you're coming from, and the concern regarding the scope and missing RFC is completely fair.

To clarify the original intent: the goal was to use a connected tracker (Trakt) as the watch state source. The proper approach would have been to open a feature request to align on the concept before submitting a PR.

Regarding AI assistance: yes, I used AI tooling throughout development. While I noted this in the Testing section, I want to be entirely transparent here. I reviewed and tested every change, but a substantial portion of the code was AI generated. Additionally, the recent support activity was just a friend helping me test the feature.

Moving forward, if you think the core idea of giving users the option to use a tracker as the watch-state source is worth exploring, I'd value your input on how it should take shape. 

Either way, based on my testing, the existing Plex → Trakt ID resolver has gaps in some cases. A smaller scoped starting point might be to improve that mapping first. Once that's reliable, revisiting the watch state idea will be much simpler.

If you don't mind, I'll open a separate issue documenting the resolver gaps I've found so we have a concrete place to start.

edde746:
Your issue will just be closed if it's slop like this response.

PR #1164, opened 2026-05-27
Add Tizen TV platform support

George-Fam:
Adds Samsung Tizen TV as a supported platform (Tizen OS 6.0+, tested on a Samsung QN43LS01B running Tizen 10).

## Overview

This PR adds a fully custom Tizen player backend written in C#/.NET (`tizen/`), wires up platform detection, subtitle rendering, and key event routing, and adds a CI job that produces a signed TPK.

## Custom player (`tizen/TizenMediaPlayer.cs`)

`video_player_tizen` package was evaluated and ruled out because it lacks audio/subtitle track enumeration, external subtitle support, and frame-accurate seeking. It also routes decoded frames through Flutter's GPU compositor via TBM surfaces, which caused severe playback issues. The custom backend bypasses all of that.

mpv is not an option, LibMPV is not available on Tizen and there are no plans to port it ([flutter-tizen/flutter-tizen#588](https://github.com/flutter-tizen/flutter-tizen/issues/588)).

Video is rendered via Tizen's hardware compositor overlay (`Display(ElmSharp.Window)`) on a dedicated hardware plane, completely separate from Flutter's render pipeline. Flutter renders a transparent hole at the video widget position; the native window sits beneath and is repositioned via `SetRoi()` on every layout change. This keeps Flutter's GPU fully available for the UI.

`BoxFit` mode cycling is forwarded to the native player via `setNativeDisplayMode()` since the Flutter widget layer has no effect on the hardware overlay.

## Tizen plugin replacements

flutter-tizen provides drop-in `*_tizen` variants of several Flutter plugins. These are added to `pubspec.yaml` alongside the originals and auto-selected by the flutter-tizen toolchain at build time:

| Replaced by                       | Original                                                      |
| --------------------------------- | ------------------------------------------------------------- |
| `shared_preferences_tizen`        | `shared_preferences`                                          |
| `url_launcher_tizen`              | `url_launcher`                                                |
| `path_provider_tizen`             | `path_provider`                                               |
| `device_info_plus_tizen`          | `device_info_plus`                                            |
| `connectivity_plus_tizen`         | `connectivity_plus`                                           |
| `package_info_plus_tizen`         | `package_info_plus`                                           |
| `wakelock_plus_tizen`             | `wakelock_plus`                                               |
| `sqflite_tizen` + `drift_sqflite` | `sqlite3_flutter_libs` (Tizen doesn't expose `libsqlite3.so`) |

`window_manager` has no Tizen variant, so `TizenWindowManager.cs` registers on the same method channel and returns TV-appropriate defaults for Tizen (always fullscreen, fixed 1920x1080, all setters are no-ops).

## Platform detection

`PlatformDetector.isTizen()` is a compile-time constant (`bool.fromEnvironment('TIZEN_BUILD')`), passed through the flutter-tizen toolchain — a runtime check isn't viable since Tizen reports `Platform.isLinux == true`. `isTV()` includes Tizen; `isDesktopOS()` excludes it.

## Event routing

When the EFL video window is shown, the Wayland compositor gives it keyboard focus and Flutter stops receiving remote key events. To fix that I had to:

- Empty Wayland input region on the video window so all normal keys (d-pad, enter) fall through to Flutter's DALi window
- KeyGrab for `XF86Back` specifically, since the back key is intercepted by the Tizen system before Wayland delivers it. The native side relays it to Flutter via the event channel, where `VideoPlayerScreen` subscribes and calls `_handleBackButton()`.

The global key handler in `key_events.dart` also had a latent bug: the Tizen navigation block was gated on `_videoPlayerNavigationEnabled`, which is always `false` on Tizen (the TV detection service only covers Android TV and Apple TV). The guard has been removed so the block actually runs. An `uiElementHasFocus` guard was added so that when the controls overlay is open and a chip or menu item has primary focus, nav keys are propagated through the focus tree instead of being consumed, enabling D-pad chip navigation and hold-timer long press in the video menus.

## Subtitles

Tizen's Capi player has no external subtitle URI support, so both paths are handled in Dart:

- Embedded subtitles via the Capi `SubtitleUpdated` callback, relayed through the event channel and rendered as a Flutter widget overlay
- External subtitles (SRT/VTT) fetched and parsed in Dart, driven by the position stream
- Secondary subtitles driven by the same Dart parsing pipeline (Capi only exposes one text track)

Subtitle style settings (font size, colour, border, bold, italic, background colour/opacity) are read from `SettingsService`. The overlay widgets live in `lib/mpv/player/platform/tizen_subtitle_overlay.dart`.

## Bootstrap changes

- Gamepad service is skipped on Tizen (`universal_gamepad` has no Tizen implementation; TV remote input comes through key events)
- Image cache budget reduced to 30 MB on Tizen
- HTTP client uses a tuned `IOClient` on Tizen (Cronet and WinHTTP are unavailable)
- Database uses `drift_sqflite` / `sqflite_tizen` on Tizen since `libsqlite3.so` is not exposed

## What isn't implemented yet

- TV remote long press on list tile menu items (secondary subtitles): `FocusableListTile` has no keyboard hold-timer path; `onLongPress` only fires on touch. Affects all TV platforms; would need another PR.

## Known permanent limitations

- No GLSL shaders:
  - Video renders on a hardware compositor outside Flutter's GPU pipeline, so there is no texture to attach a shader to.
- No audio delay:
  - The Capi player manages audio/video sync internally and exposes no API to offset audio timing.
- Subtitle delay is positive-only for embedded tracks:
  - Dart-parsed subtitle tracks support both positive and negative delay.
  - Embedded tracks (driven by the Capi callback) only support positive delay. Negative delay would
    require buffering future cues before they're received, which Capi doesn't allow.
- Subtitle track labels show language only (`EN` instead of `EN - SRT`):
  - Capi only exposes the language code per track; showing the codec format (SRT, ASS, etc.) would
    require either probing the container with ffprobe (no Tizen package available) or maintaining a separate track metadata lookup, neither of which is currently implemented.

## CI setup

A `build-tizen` job was added to `.github/workflows/build.yml`. It installs flutter-tizen 3.41.9, Tizen Studio 6.0, and the required toolchain, then builds and signs a TPK.

Four repository secrets are required.

| Secret                       | Description                                                        |
| ---------------------------- | ------------------------------------------------------------------ |
| `TIZEN_AUTHOR_CERT_BASE64`   | Author `.p12` certificate, base64-encoded (`base64 -i author.p12`) |
| `TIZEN_AUTHOR_CERT_PASSWORD` | Password for the author certificate                                |
| `TIZEN_DIST_CERT_BASE64`     | Distributor `.p12` certificate, base64-encoded                     |
| `TIZEN_DIST_CERT_PASSWORD`   | Password for the distributor certificate                           |

Both certificates are issued through Tizen Studio (Certificate Manager). The author cert identifies the developer; the distributor cert is required for installation on a real device.

## Cross-platform impact

**SDK constraint downgrade.** flutter-tizen currently only supports up to Flutter 3.41.9, so `pubspec.yaml` was rolled back from `>=3.44.0` / Dart `>=3.12.0` to `>=3.41.9` / Dart `>=3.11.0`. Some Flutter 3.44 and Dart 3.12 language features already in use had to be reverted until flutter-tizen updates to 3.44 (https://github.com/edde746/plezy/commit/209101796ba186191116ac0de1c79b6cbb8273f5). Non-Tizen CI still runs against the latest Flutter.

**macOS plugin registrant.** Adding `drift_sqflite` pulls `sqflite_darwin` into `macos/Flutter/GeneratedPluginRegistrant.swift` as a side effect. There is no symbol conflict and it is never called on macOS, but it can't be cleanly avoided without platform-conditional dependency support in pub.


edde746:
AI used? If so, which?

George-Fam:
Yes, I used claude as coding/review assitant for this PR, but I've personally reviewed and  tested it on my Samsung TV as well.

Sorry for the big PR.

edde746:
I think this is better suited as a fork, I don't really want to maintain a Tizen version and I don't think I could ever publish it due to Samsung's requirements to do so. It also has significant impact on the other platforms by blocking engine upgrades.

balcsida:
@George-Fam — an update you deserve to hear, since this PR is sitting closed while your work is very much alive.

I picked the Tizen port back up on a fork (`balcsida/plezy`, branch [`feat/tizen-6-tv`](https://github.com/balcsida/plezy/tree/feat/tizen-6-tv)), and a lot of it stands on what you built here. **Jellyfin playback now runs on a retail set — Samsung UE55AU7022KXXH, Tizen 6.0, under the real `com.edde746.plezy` identity: picture, audio, seek and the Flutter controls together**, installed as an in-place signed upgrade from a CI-built, author-signed TPK.

Things taken directly from your PR:

- The platform detection. Compile-time `bool.fromEnvironment('TIZEN_BUILD')`, with `isTV()` including Tizen and `isDesktopOS()` excluding it. Your note that a runtime check isn't viable because Tizen reports `Platform.isLinux == true` saved the obvious dead end.
- `universal_gamepad` has no Tizen implementation and the remote arrives as key events, so the gamepad service stays unstarted.
- `os_media_controls` likewise — every call was a caught-and-logged `MissingPluginException`, once a second during playback.
- The EFL/Wayland focus problem and the `XF86Back` `KeyGrab`. That one would have cost days to rediscover.

**The objection that closed this PR appears to be gone.** The SDK downgrade was the blocker — flutter-tizen pinning 3.41.9 and dragging the floor back. The fork branch keeps `flutter: ">=3.47.0"` / `sdk: ">=3.12.0"`, byte-identical to upstream `main`, because flutter-tizen has moved on. I ran upstream's own suite against the branch: 7,449 tests with a single failure that reproduces on clean `main`, and the only analyzer findings are pre-existing ones in vendored `packages/wakelock_plus`.

Being straight about what is still open: independent cryptographic verification and Samsung Store entitlement, most of the device-matrix rows (physical remote, focus, lifecycle, long-run stability), and the six `flutter-tizen/plugins` git-pinned dependencies, which are probably the biggest remaining adoption question. The port is not production-ready and the branch's own `docs/tizen/ACCEPTANCE.md` says so.

None of that changes the main point: you got this working against real hardware first, including the custom C#/Capi backend when libmpv was never an option, and you documented the permanent limitations honestly instead of papering over them. Thank you — it made the difference between a port and a guess.

Happy to compare notes, and equally happy for you to take any of it back into your own branch.

PR #1190, opened 2026-05-29
Feat/random episode downloads

Darkmadda:
Changes:

FEAT:
-new option to download menu to allow you to download random episodes instead of just the next X episodes (Usefull when you want to get episode of already watched show for rewatch randomly)
-added "Custom amount" to tv show download , previous option only downloaded unwatched, this new feature will down any episode regardless of watched status.
-added "Selected Season" option to download allowing user to download X episodes only for specific seasons. 
-sync rules follow the above options if defined when sync is added.
-Added ability to play random episodes from selected season instead of just the entire show.
-added manual sync button to the downloads section allowing you to trigger a sync that re-evaluates download rules and processes the new deletes/downloads to keep things synced (It was only running this automatically on rerun/connection restarted) Only shows up when a sync rule exists
--- general new features note ---
for all of the new features i tried to hide any options that are not contextually needed. ie if a button has no acual application for the feature it won't show.

BUG FIX:
-fixed bug where progress circle on downloading of episodes assumed when downloading started that 100% progress was all the episode in the show and not the amount of episodes currently being downloaded/synced. the previous behavior made the progress bar useless if you were say downloading just 5 epides of something that had like 50 episodes it would only get as far as 10% filled).
-fixed bug where buttons while viewing a show would overflow the viewport on small screens and would make buttons at the end not accessible since they were off screen. I made them hide buttons if there wasn't room and made sure what was hidden was in the 3 dot menu popup so that no functionality would be lost.
-fixed but where if you had downloaded episodes and then deleted some the show would no longer be found when you browsed your library (You could still search for it but it disappeared from you show list).
-fixed bug where if episodes are downloaded clicking the download button again it doesn't prompt for what you want but instead just queues up and downloads the entire show.

OTHER:
fixed several CI issues that came back from flutter analyze
unit tests added for fixes and new features

Tested on iOS, MacOS and Android

ANOTHER NOTE REQUEST:
I couldn't find anyway to communicate with other developers to discuss what bugs i find and want to fix and features i think need to be added. Usually with projects a forum or discord server is setup so developers can coordinate. If there is something and i just couldn't find please let me know. if there isn't anything out there someone set it up and let me know. I don't like contributing blindly without a plan


edde746:
A lot of changes in one PR is not great. Please split this into multiple PRs and clean up the commits.

> couldn't find anyway to communicate [...]

Use the GitHub issues.

PR #1196, opened 2026-05-30
fix(media-detail): collapse action buttons that overflow small screens

Darkmadda:
## Problem
On small-screen phones, the detail-screen action row rendered as a plain `Row`, so the trailing buttons — including the three-dot overflow menu — ran past the viewport and were unreachable.

## Fix
Wrap the action row in a `LayoutBuilder` and greedily drop buttons that don't fit, in priority order:

**Trailer → Mark Watched → Shuffle → Download**

- **Play** and the **⋮ More** menu are pinned (never dropped) — the menu is the fallback that keeps hidden actions reachable.
- A button is only dropped when it stays reachable via the menu. **Offline mode** has no ⋮ menu, so Mark Watched is pinned there too (it queues watch state for sync).
- Drop logic is **mobile-only** — skipped on TV to avoid stranding D-pad focus on a vanished button.
- The Play button's dynamic, localized label is measured with a `TextPainter`; icon buttons are fixed-size. Single layout pass, no flicker.
- `LayoutBuilder` re-runs on rotation/resize and when the trailer button appears after Plex extras load.

A **"Play trailer"** item was added to the three-dot menu (via a new optional `onPlayTrailer` callback on `MediaContextMenu`) so the trailer stays accessible when its row button is hidden. The item only appears on the detail screen (card/browse menus don't pass the callback), and reuses the existing `tooltips.playTrailer` string — no new i18n.

## Scope
Applies to all media detail types (movie/show/season/episode) using the shared action row, since the overflow is layout-driven, not content-specific.

## Testing
Verified manually on a small viewport: confirmed the drop order, that the ⋮ menu stays on-screen, that "Play trailer" works from the menu, that offline keeps Mark Watched visible, and that TV is unchanged.


edde746:
Closed in favor of https://github.com/edde746/plezy/commit/f5012b36ea7ecb7201fc885dff3753de2f08c226

PR #1239, opened 2026-06-04
fix(jellyfin): use defaltStreamIndex to choose server selected tracks

a-collado:
Fix how the audio and subtitle server-selected tracks are chosen when using a Jellyfin server. 

If the backend is Jellyfin , and the media info includes `defaultAudioStreamIndex` or `defaultSubtitleStreamIndex`, the index is used to match the track with one of the available player tracks. 

Jellyfin uses this index to expose the selected track, so not using it makes the player not respect the server-selected tracks. E.g., if you start watching a movie in the Jellyfin web UI and manually change the audio or subtitle track, without this change , these changes would not reflect in Plezy if you continue watching the movie there.

<h2>Testing</h2>

1. `flutter analyze` on the changed file: no issues.
2. `flutter test test/services/track_manager_test.dart`: all tests passed correctly
3. Manually change a track in the Jellyfin web UI (you have to start playing the media for the changes to be applied) and then play the media in plezy, it should automatically select the same tracks you selected on the Web UI.

### DIFF
--- a/lib/services/track_selection_service.dart
+++ b/lib/services/track_selection_service.dart
@@ -624,7 +624,7 @@ class TrackSelectionService {
 
   /// Select the best audio track based on priority:
   /// Priority 1: Preferred track from navigation
-  /// Priority 2: Plex-selected track from media info
+  /// Priority 2: Server-selected track from media info
   /// Priority 3: Per-media language preference
   /// Priority 4: User profile preferences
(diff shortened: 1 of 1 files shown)


edde746:
Closing in favor of https://github.com/edde746/plezy/commit/79c06b82d303c704ec61b068097d7b0c3a68c6de

PR #1378, opened 2026-06-18
feat(linux): Update layout to support Linux Mobile

radmorecameron:
I'm running Plezy on a Linux Mobile device and this change makes the experience much more pleasant without the side bar :smile: 

Before:
Plezy with a side bar feeling a bit clunky at mobile width

After:
screenshot of Plezy on a linux mobile device displaying nicely

### DIFF
--- a/lib/utils/platform_detector.dart
+++ b/lib/utils/platform_detector.dart
@@ -235,14 +235,15 @@ class PlatformDetector {
     return isDesktop(context) || isTV();
   }
 
-  /// Detects if running on a mobile platform (iOS or Android).
+  /// Detects if running on a mobile platform (iOS, Android or Linux Mobile).
   /// Excludes TV platforms (Android TV / Apple TV) even though the underlying
   /// OS is iOS or Android.
   /// Uses Theme for consistent platform detection across the app.
   static bool isMobile(BuildContext context) {
     if (isTV()) return false;
     final platform = Theme.of(context).platform;
-    return platform == TargetPlatform.iOS || platform == TargetPlatform.android;
+    return platform == TargetPlatform.iOS || platform == TargetPlatform.android ||
+      (platform == TargetPlatform.linux && getDiagonalInches(context) <= 7.0);
   }
 
   static bool isHandheld(BuildContext context) {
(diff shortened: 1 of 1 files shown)


edde746:
What about Linux desktops with small screens? Is there maybe a better way to actually detect mobile Linux environments?

radmorecameron:
Screen size is probably the best way since the phone could also be plugged into a monitor (in which case it will behave like the desktop app).

If the 13-inches diagonal seems like too large, I'd be happy to lower it as well. I chose that size because it seems to be a tablet size but I imagine the side bar would be more usable on a tablet sized screen

radmorecameron:
I've decreased the check from `13.0` to `7.0` so small desktop monitors should behave the same.

Tablets and open foldables will show the side bar instead of the bottom nav bar

radmorecameron:
Do you think this change would be mergable with the decreased size change (7.0 instead of 13.0)? @edde746 

I've been running a build with these changes somewhat infrequently and have found the changes provided some nice QOL improvements on Linux Mobile 🙂

edde746:
It's not about the size of the screen, I feel like this is the wrong approach to detect mobile Linux.

radmorecameron:
I think any other approach would not work well if someone plugs their phone into a monitor :thinking: do you have any other ideas on how to approach this?

edde746:
Physical diagonal is the part I don't want to merge: it silently reclassifies small/high-DPI desktop panels and VMs reporting odd EDID data as phones, and there's no threshold that's right for both a 7" handheld and a 10" desktop panel.

Two directions I'd look at instead (speculation — I haven't verified either, that's yours to validate):

1. Detect mobile Linux from the session, not the screen: `XDG_CURRENT_DESKTOP` / `XDG_SESSION_DESKTOP` (Phosh, Plasma Mobile), or tablet-mode/touch signals (DBus tablet-mode, libinput touchscreen presence). That answers "is this mobile Linux" honestly instead of inferring it from geometry.
2. Split the two questions. The docked-phone case is an argument *for* that split, not against env detection: let the env check gate mobile-Linux eligibility, and let the actual sidebar-vs-bottom-nav choice follow logical window width like the app's existing responsive splits. Docked phone → bigger window → desktop layout, undocked → mobile layout, and desktop Linux is never affected regardless of monitor size.

Happy to review an approach along those lines.

radmorecameron:
Hmm, I see what you mean about the DPI with the diagonal calculating.

- `XDG_CURRENT_DESKTOP` would work for Phosh but not any other DE so I don't think that is the right approach.
- tablet-mode/touch signals (DBus tablet-mode, libinput touchscreen presence) wouldn't work well in a sandbox (like flatpak) so I don't think it would be good to use that either

Instead of using the diagonal and guessing the DPi, do you think just using `MediaQuery.sizeOf` directly would work (ex: make sure smallest size is less than 600px to determine if it should use a mobile friendly layout). (Edit: this seems to be more-or-less how other Linux apps adapt to mobile sizes)

PR #1424, opened 2026-06-26
fix: coalesce held timeline seeks on direct play

ctismer:
On current hardware you won't notice this, but on weaker devices (I hit it on a
Fire TV Stick 4K) the rendering can't keep up and the buffering gets tangled —
a spinning wheel that never clears, recovered only by leaving and re-entering
playback.

Holding fast-forward fires a real seek on every key-repeat; on a direct stream
each one re-issues an HTTP range request until ExoPlayer wedges in
STATE_BUFFERING. The timeline already coalesces held seeks for transcodes — this
routes direct play through that same preview/debounce path, committing a single
seek on release. No new settings, no native changes.

Disclosure: written with help from an AI coding assistant.

### DIFF
--- a/lib/widgets/video_controls/desktop_video_controls.dart
+++ b/lib/widgets/video_controls/desktop_video_controls.dart
@@ -533,15 +533,13 @@ class DesktopVideoControlsState extends State<DesktopVideoControls> {
   KeyEventResult _handleTimelineKeyEvent(FocusNode _, KeyEvent event) {
     final key = event.logicalKey;
 
-    // Handle key release to reset progressive seek state
+    // Handle key release: commit the accumulated scrub target with a single
+    // seek (a no-op when nothing is pending) and reset progressive seek state.
     if (event is KeyUpEvent) {
       if (key == LogicalKeyboardKey.arrowLeft || key == LogicalKeyboardKey.arrowRight) {
-        if (_trackControlsState.isTranscoding) {
-          _flushTimelinePreviewSeek();
-          _resetSeekState();
-          return KeyEventResult.handled;
-        }
+        _flushTimelinePreviewSeek();
         _resetSeekState();
+        return KeyEventResult.handled;
       }
       return KeyEventResult.ignored;
     }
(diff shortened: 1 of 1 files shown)


edde746:
Manually tested?

ctismer:
Yes :) This was the show-stopper that made Plezy unusable on our old Fire TV Stick 4K. I'm comparing Plezy, Wholphin, Moonfin and Jellyfin side by side on that stick — they all struggle on weak hardware in their own ways, and I'm fixing each so I can pick the best one. Held fast-forward used to flood real seeks until the player got stuck in buffering; with this change it's smooth.

micahmo:
@ctismer Is this the issue where playback will randomly jump to a different place in the timeline when it's struggling to buffer?

ctismer:
Not sure it's the same one. This PR fixes a specific case: on direct play, holding the DPAD fast-forward/rewind fired one real seek per key-repeat, which flooded the player until it got stuck buffering (spinning wheel that only recovered by leaving and re-entering). The fix coalesces the held seeks so a single seek commits on release.

edde746:
Thanks @ctismer — the approach here is right, and it matches what the media3 ecosystem does for held seeks (media3's own `DefaultTimeBar` commits via `onScrubStop` ~1s after the last key event, Leanback's transport glue is preview-then-commit, and Jellyfin Android TV accumulates skips behind an 800ms debounce). Normal-mode ExoPlayer has no internal seek coalescing, so doing it in the UI is the correct place.

Landed your commit on main via cherry-pick with authorship preserved (ed843b4b), plus three follow-up hardening commits found in review:

- 632d66f6 clears the timeline-seek dedup state per gesture. Routing key-seeks through `onSeekEnd` made them subject to the drag-path duplicate check, so a coalesced flush landing on a stale drag target (realistic at the 0/duration clamp boundaries) skipped the real seek while still reporting completion.
- 3cad9173 raises the scrub debounce 350ms → 800ms so it sits above the initial key-repeat delay (~400–500ms on Android/TV) — otherwise every hold still committed one extra mid-hold seek before repeats began. Release flushes synchronously, so taps are unaffected.
- f9abcd44 holds the timeline preview until playback settles near the committed target (10s ceiling). On a still-buffering device the unconditional 2s clear let the next press re-base off the stale live position and re-target the same spot — exactly the slow hardware this fix targets.

Closed automatically since the change is now on main.

PR #1428, opened 2026-06-27
feat: add client-side Watchlist with bookmarking, grouped grid UI, and navigation integration

BlazzzPlay:
## Summary

Adds a client-side Watchlist feature — a local bookmark system for movies, shows, seasons, and episodes. Profile-scoped, offline-capable, and purely additive (no existing behavior changes).

## What's included

- **Drift table** WatchlistItems with composite PK (profileId, globalKey) — schema v16→v17 migration
- **WatchlistProvider** — ChangeNotifierProxyProvider with stream-based reactive state
- **WatchlistScreen** — grouped grid by kind (Movies/TV Shows/Seasons/Episodes), empty state, clear-all with confirmation
- **Bookmark toggle** — FocusableAction in media detail screen action buttons
- **Navigation integration** — bottom nav tab, side rail entry, companion remote via next/prev
- **i18n** — full key set across 16 locales

## Changes

- 15 conventional commits across 5 layered PR slices
- 4 new source files, 8 modified source files
- 5 new test files (63 new tests)
- 2543 total tests passing, zero regressions

## Test plan

- `flutter test` — 2543/2543 passing
- `flutter analyze` — clean on all changed files
- APK built and tested on Android device — watchlist works end-to-end

## Notes

Port of the toffbrawny/plezy watchlist implementation, adapted and modernized for current upstream codebase (~2300 commits ahead of the original fork).

### DIFF
--- a/lib/database/app_database.dart
+++ b/lib/database/app_database.dart
@@ -47,6 +47,7 @@ enum OfflineActionType {
     ApiCache,
     OfflineWatchProgress,
     SyncRules,
+    WatchlistItems,
     Connections,
     Profiles,
     ProfileConnections,
@@ -61,7 +62,7 @@ class AppDatabase extends _$AppDatabase {
   AppDatabase.forTesting(super.e);
 
   @override
-  int get schemaVersion => 16;
+  int get schemaVersion => 17;
 
   @override
   MigrationStrategy get migration {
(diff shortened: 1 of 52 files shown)


edde746:
I don't think a local watchlist is the right direction as users would most likely want to sync with other devices. I think it'd be better to "piggyback" of an existing service(s) like Seerr or Trakt.

Please create an issue/suggest your solution in an issue when it comes to large changes like these before implementing to save everyone (especially yourself) time and effort.

BlazzzPlay:
Thanks for taking the time to review this, and for the honest feedback. You are right — I should have opened an issue first to discuss the approach before jumping into implementation. Lesson learned.

The direction of piggybacking on Trakt or Seerr makes total sense for multi-device users. I will open an issue to discuss that approach when I have time to explore it properly.

Appreciate your time and the work you put into Plezy. 🙏

PR #1429, opened 2026-06-27
fix: make crash reporting opt-in instead of opt-out (privacy concern)

BlazzzPlay:
## Problem

Crash reporting (Sentry) is enabled by default when `ENABLE_SENTRY` is compiled in. The `crashReporting` `BoolPref` defaults to `true`, meaning crash data is sent to a remote server (`bugs.plezy.app`) **without explicit user consent**.

This is a privacy concern because:
- Users have no initial consent dialog — they must discover and disable it themselves in Settings > Advanced
- The DSN points to a self-hosted Sentry instance, which users may not expect or trust
- Crash reports may contain sensitive data (stack traces, device info, usage patterns)

## Two-layer control system (for context)

| Layer | Default | Mechanism |
|---|---|---|
| Compile-time (`ENABLE_SENTRY`) | `false` | `--dart-define` at build time |
| Runtime (`crashReporting`) | `true` → **`false` (this PR)** | `BoolPref` in SharedPreferences |

The runtime toggle in Settings already exists — this PR only changes the **default** from opt-out to opt-in. Users who want to contribute crash reports can still enable it explicitly.

## Changes

- `lib/services/settings_service.dart`: change `crashReporting` default from `true` to `false`
- `test/services/settings_service_test.dart`: update test to reflect new default and keep `refreshListenables` test meaningful

## UX behavior after fix

1. Fresh install → crash reporting is **off** by default
2. User goes to Settings > Advanced → toggles "Crash Reporting" **on** if they want to help
3. Existing users with the setting already stored are unaffected (SharedPreferences persists their choice)

### DIFF
--- a/lib/services/settings_service.dart
+++ b/lib/services/settings_service.dart
@@ -331,7 +331,7 @@ class SettingsService extends BaseSharedPreferencesService {
   static const String defaultCreditsPattern = r'(?:^|\b)(?:outro|closing|credits?|ending)(?:\b|$)|^ed(?:\s?\d+)?$';
 
   static const enableDebugLogging = BoolPref('enable_debug_logging', onWrite: setLoggerLevel);
-  static const crashReporting = BoolPref('crash_reporting', defaultValue: true);
+  static const crashReporting = BoolPref('crash_reporting', defaultValue: false);
   static const enableHardwareDecoding = BoolPref('enable_hardware_decoding', defaultValue: true);
   static const enableHDR = BoolPref('enable_hdr', defaultValue: true);
   static const preferredVideoCodec = StringPref('preferred_video_codec', defaultValue: 'auto');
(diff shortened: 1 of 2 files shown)


edde746:
Stack traces are not sensitive data. No usage patterns are identifiable from the reports. None of the information sent is personally identifiable and it helps improve the app, very few would manually opt in, making it entirely useless.

BlazzzPlay:
Unrelated to this PR, but since issues are disabled here — have you considered adding Chromecast or DLNA streaming support?

The main use case would be LG and Samsung TVs that run WebOS/Tizen and can't install Plezy directly. A Chromecast sender would cover the widest range of devices without needing native TV ports.

Is casting anywhere on the roadmap, or is it intentionally out of scope?

edde746:
Issues are not disabled...

PR #1449, opened 2026-06-29
Add Letterboxd search button from movie summary screen

spaciousmind:
Adds a shortcut button to open Letterboxd search for the current movie using title and year.

Tested locally and used during normal browsing.


edde746:
1. It's not localized
2. It won't work on devices without a browser (tvOS and some Android TV)

Most importantly, I think features like this should be avoided as it sets a precedence of adding links and integrations to every possible service and API leading to clutter and UX complexity.

I don't know how the Letterboxd API works, but the cleaner solution would be to add it as a tracker (like we have Trakt, SIMKL, MAL and AniList).

PR #1459, opened 2026-06-30
MacOS Video Optimisations for Older Machines

tanenyi:
Adds two macOS-only settings under Settings > Video Playback to reduce CPU/GPU load and fan noise on older Macs:

## Renderer Pipeline

Lets users switch the mpv video output backend between:
- **gpu-next + Vulkan** (default) -- best HDR/tone-mapping, heaviest
- **gpu + Vulkan** -- simpler shaders, still uses MoltenVK
- **gpu + OpenGL** -- bypasses MoltenVK entirely, lightest

Takes effect on the next video played.

## Flutter Renderer (Impeller/Skia)

Toggle to use Skia instead of Impeller (Metal) for Flutter's UI rendering. This can significantly reduce idle GPU usage on machines where Impeller's Metal backend runs hot.

Modifies the app bundle's `FLTEnableImpeller` Info.plist key. Requires app restart.

## Notes

- Both settings are macOS-only
- The Impeller toggle only works on unsigned/dev builds (modifies Info.plist on disk)
- Renderer pipeline properties are set at mpv init time and can't change mid-playback

### DIFF
--- a/lib/mpv/player/player.dart
+++ b/lib/mpv/player/player.dart
@@ -348,8 +348,7 @@ abstract class Player {
   /// - true: Use ExoPlayer (default, better hardware support)
   /// - false: Use MPV (more features, ASS subtitle rendering)
   ///
-  /// [rendererConfig] is an optional map of renderer properties (vo, gpuApi,
-  /// gpuContext) passed to the native layer at initialization time (macOS only).
+  /// [rendererConfig] overrides mpv renderer properties (macOS only).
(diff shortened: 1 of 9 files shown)


edde746:
Have you manually tested this? Do you have performance measurements?

tanenyi:
yes, manually tested and the fans on my MacBookPro15,1 doesn't spin up as quickly as before. unfortunately I don't have performance measurements

Issue #1009, opened 2026-05-11
Display show/movie logo on player details screen


PrebbLord:
### Problem Statement

The show name and episode title are quite small on the details/player screen. Especially when viewing on a TV.

### Proposed Solution

I propose that the show/movie's logo appears, along with a larger episode title.

<img width="1650" height="924" alt="Image" src="https://github.com/user-attachments/assets/d74df04e-ce92-42b2-9ebb-348975ec4af4" />

### Target Platform

All platforms

### Media Server

Both

edde746:
Doesn't really work since the back button needs to exist on desktop.

PrebbLord:
Oh yeah, imagine a back button in my image

edde746:
I did, it doesn't really work, alignment would be weird.

PrebbLord:
How about this?

<img width="1646" height="921" alt="Image" src="https://github.com/user-attachments/assets/55f0ae3c-2936-4c41-b6df-06ab9701ce0b" />

PrebbLord:
Or like this

<img width="1647" height="925" alt="Image" src="https://github.com/user-attachments/assets/7fdbe831-4aab-46df-9f3a-c2c358d6f78e" />

PR #1460, opened 2026-06-30
Show title treatment logo in video player header

tanenyi:
When a movie or series has a clear logo image (title treatment), display it in the video player header next to the back button instead of plain text — similar to how streaming apps like Fladder/MUBI show decorative title art during playback.

## Changes

VideoControlsHeader now checks metadata.clearLogoPath and renders the logo via CachedNetworkImage when available
Uses MediaImageHelper for DPR-aware optimized image URLs (same pattern as TV spotlight and detail screens)
Gracefully falls back to the existing text title when:
- No clear logo is available
- The image URL is empty (e.g. offline with no cache)
- The image fails to load
- Shows text title as placeholder while the logo loads

## SS
Screenshot 2026-07-01 at 9 42 00 AM
Screenshot 2026-07-01 at 9 42 26 AM
Screenshot 2026-07-01 at 9 42 59 AM

### DIFF
--- a/lib/widgets/video_controls/widgets/video_controls_header.dart
+++ b/lib/widgets/video_controls/widgets/video_controls_header.dart
@@ -1,9 +1,15 @@
+import 'package:cached_network_image_ce/cached_network_image.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 import 'package:plezy/utils/formatters.dart';
 
 import '../../../media/media_item.dart';
+import '../../../media/ids.dart';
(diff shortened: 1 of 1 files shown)


edde746:
When making UI changes, please attach screenshots.

tanenyi:
added ss 👍

edde746:
While I think the addition is nice in theory, the legibility is bad due to the size. You might want to take a look at #1009, however, those designs also have some issues on some devices, which is why I have not implemented it.

Issue #749, opened 2026-03-25
Fire tv cube - constant freezes


xaelus:
### Bug Description

Often times when in watch together with my partner on her Mac.  When they join, oftentimes the entire app will cause my fire tv to lockup.  Sometimes I get to library screen select something to play and it hard locks.  

Happening frequently after most recent updates this past week.  

### Steps to Reproduce

Watch together
Partner joins freezes entire os

Watch together
Joins
Select item to play
Locks os requiring hard reboot

### Expected Behavior

Not freezing and requiring hard reset.  

### Actual Behavior

Complete freeze.  No buttons respond only solution is hard reboot of fire tv.  

### Plezy Version

1.29.0

### Platform

Android

### Device Information

Fire tv cube newest gen

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

edde746:
Could you try [this build](https://github.com/edde746/plezy/actions/runs/23537696626/artifacts/6099777758)?

xaelus:
I can’t seem to sideload this.  I tried the v7 and arm64.  Had to unzip a few times till all.  Both were recognized but days app installed.  Tried downloaded and es file.   Also of note I noticed tremendous slowdown on menus, thumbnails.  The iPad is instant but fire tv got super slow.

edde746:
> days app installed

What?

And yeah, any iPad will be more performant, this is because Android in general just isn't performant and running it on the worst chips possible won't help.

xaelus:
Sorry must have stroked out.  I have the original installed, and when trying to trigger either version of apk it attempts then says not installed.  Should I uninstall the try apk?

edde746:
Yes, uninstall then install the apk.

Some GeekBench scores for fun:
| SoC / Device Class                        |  GB5 Single |     GB5 Multi | POP1-G vs     |
| ----------------------------------------- | ----------: | ------------: | ------------- |
| **Amlogic POP1-G** (Fire TV Cube 3rd gen) | **140–146** |   **490–524** | baseline      |
| **Apple A9** (iPhone 6s / iPad 5)         | **522–562** | **1035–1050** | 0.26× / 0.49× |
| **Apple A9X** (iPad Pro 9.7)              |     **639** |      **1124** | 0.23× / 0.45× |
| **Apple A10 Fusion** (iPhone 7 / iPad 7)  | **756–761** | **1341–1413** | 0.19× / 0.37× |

[...]

PR #1479, opened 2026-07-02
Optimize TV layout with frustum culling, Vulkan (Impeller) on Android 11+, and memory limits

KakarottoCake:
### Overview
This Pull Request introduces rendering and layout optimizations to make the TV user interface run smoothly (specifically targeting 120Hz refresh rates on modern TV devices like the Google Streamer 4K). 

It implements frustum culling on scrollable rows, enables Vulkan (Impeller) rendering on supported Android TV hardware, and optimizes image cache budgets to prevent out-of-memory crashes.

---

#### 1. Vertical & Horizontal Frustum Culling
*   **Vertical Row Culling**: Tracks the scroll offset in the browse rail and replaces off-screen rows with lightweight, stable-height `SizedBox` placeholders. This completely bypasses the CPU/GPU layout and image loading costs of off-screen rows while keeping D-pad navigation stable.
*   **Horizontal Card Culling**: Enforces a low `scrollCacheExtent` (using the modern `ScrollCacheExtent.pixels` API) on horizontal list views so off-screen cards are immediately recycled and disposed of.
*   **Pre-rendering Budget**: Restricts the vertical caching on TV platforms to prevent pre-rendering too many rows off-screen.

#### 2. Vulkan (Impeller) Enablement on Modern TVs
*   Lifts the blanket forced-Skia (OpenGL) fallback for all Android TV devices in the native `MainActivity.kt`.
*   TV boxes running Android 11 (API 30) or higher (like the Google Streamer 4K) will now render using **Impeller (Vulkan)**. This eliminates Skia's runtime shader compilation jank.
*   Retains the OpenGL/Skia fallback for legacy TV hardware (pre-Android 11) and devices with driver limitations.

#### 3. TV Memory & VRAM Optimizations
*   Introduced a TV-specific image cache limit in `device_performance.dart`, capping memory usage at `64MB` (down from 100MB) and a maximum of `500` images.
*   This frees up VRAM for Vulkan render swapchains and native 4K video decoders.

#### 4. Developer Build Convenience
*   Updates `build.gradle.kts` to automatically fall back to the standard `debug` signing key for release builds if `key.properties` is missing, ensuring builds remain installable on test devices.


edde746:
Thanks for digging into TV performance — I went through each of the ideas here in detail. Two of them landed on main (one under your authorship), the others I have to decline, with reasoning below.

**Landed:**

- **TV image cache cap** — cherry-picked with you as author in cd8212fb. Good catch; TV boxes really do fight the 4K decoder for RAM, and this class of device has produced OOM reports before.
- **Impeller on Android TV** — landed in 7b1c150e, but with a different gate than API 30. The blanket disable wasn't arbitrary: it shipped in e759dafa and was validated by #749, where a Fire TV Cube 3rd gen froze hard until a Skia build fixed it. The problem with gating on API 30 is that Fire OS 8 *is* Android 11 (API 30), so your check re-enables Impeller on exactly the hardware that motivated the disable. The gate that landed instead: Android 12+, `FEATURE_VULKAN_HARDWARE_VERSION` ≥ 1.1, and not Amazon-manufactured — which still covers the Google TV Streamer 4K you're targeting. Sentry events now carry a `renderer` tag so we can widen (or narrow) this from real crash data.

**Declined:**

- **Vertical row culling** — as written it's unreachable: the vertical rail is a lazy `ListView.builder` with `itemExtentBuilder`, so Flutter only ever builds rows within the viewport plus the 250px default cache extent. Your cull margin is 1.2× viewport height (~650+ logical px), strictly wider than the build window, so `isVisible` is true for every row that ever gets built and the `SizedBox` placeholder branch never executes. Meanwhile the per-item `ValueListenableBuilder` on the scroll offset rebuilds every built row on every frame of a scroll animation — the file's current design deliberately keeps row content as stable child subtrees to avoid exactly that.
- **`ScrollCacheExtent.pixels(20)` on horizontal rails** — 20px means a card is disposed the moment it leaves the viewport and rebuilt (including image decode) as soon as it re-enters, which turns into pop-in and scroll-time CPU churn — the opposite of a 120Hz win. It's also applied unconditionally in `hub_section.dart` and `media_detail_screen.dart`, so phone/tablet fling scrolling would regress too.
- **Debug-signing fallback in `build.gradle.kts`** — harmless, but CI provisions `key.properties` from secrets and I'd rather local release builds without a keystore stay unsigned than silently debug-signed.

One housekeeping note for future PRs: 25 of the 32 changed files here are regenerated `*.g.dart` outputs from a different build_runner version (e.g. `app_database.g.dart` at +909/−2541), which made the diff mostly noise — please leave generated files out unless the generating source changed.

Closing since the substantive parts are on main. Thanks again — the memory-budget angle in particular was a genuinely useful find.

KakarottoCake:
Genuinely useful information on those rejections thanks for sharing. Still new to coding so these things excite me. Appreciate authorship and hope I can continue to find new things to help contribute.

PR #1514, opened 2026-07-09
fix(tv): don't leave the on-screen keyboard stuck on companion-remote search

l3gitpanda:
AI Used for commit. Also unable to test due to no dev account or mac.

## Summary

Searching from the **Plezy companion remote** left the custom TV on-screen keyboard (OSK) stuck on screen, covering the search results. This PR makes a remote-submitted query behave like a *submit* (run the search, focus the results) instead of *focus-to-type*, so the OSK is never left open.

## The bug

On TV, when you triggered a search from the phone companion remote:

1. The remote routed the app to the **Search** tab, which focuses the search field.
2. Focusing the field **auto-opens** the custom TV OSK (that's the intended behavior when a user navigates to search to type).
3. The query text was then injected into the field — but nothing ever dismissed the keyboard.

The OSK just sat there over the results. Its normal dismissal paths don't apply to a remote search: blur is intentionally a no-op (so the keyboard survives rebuilds/focus swaps), and the remote never sends a `Done`/`Back` key. The user already typed the query on their phone, so the OSK shouldn't have opened at all.

## Root cause

A remote search reused the same code path as "user tapped the Search tab to type": route to Search → focus input (→ auto-open OSK) → set text. That path has no step that closes the keyboard, because for interactive typing you *want* it open. There was no way to say "I have a complete query already — just run it and show results."

## The fix

Treat a query-bearing remote search as a submit, distinct from focus-to-type:

- **`SearchInputFocusable.submitSearchQuery(query)`** — new mixin method (`lib/mixins/refreshable.dart`). `SearchScreen` implements it to set the text, dismiss any open OSK, run the search immediately, and land focus on the first result when results arrive — never opening or re-opening the OSK. `setSearchQuery` keeps its old "just set text, stay focused for typing" semantics.
- **`main_screen.dart`** — the companion-remote `onSearchAction` now calls `submitSearchQuery` and routes to the Search tab with `focusSearchInput: false` for a query-bearing search, so `_selectTab` no longer focuses the input (which is what auto-opened the OSK). An empty remote search still focuses the input as before.
- **`TvKeyboardController`** — new optional, additive imperative handle on `FocusableTextField` / `FocusableTextFormField` (`lib/focus/focusable_text_field.dart`). It exposes `closeKeyboard()` (dismiss an already-open OSK and suppress auto-reopen while the field keeps focus) and `focusInputWithoutKeyboard()` (focus the field without opening the OSK). `closeKeyboard()` is needed because the phone's "Search tab" button opens the keyboard *before* the query is sent.
- **Empty-results fallback** — a remote submit that returns no matches focuses the (visible, query-filled) input *without* reopening the OSK, so the remote isn't stranded on the now-hidden previous tab.

The `FocusableTextField` change is purely additive — an optional nullable `tvKeyboardController` that defaults to `null` — so every other text field in the app is unchanged.

## Testing

Per `CONTRIBUTING.md`:

- `dart format` — no changes on the touched files.
- `flutter analyze` on the changed files — **No issues found**.
- `flutter test` on `test/mixins/refreshable_test.dart` and `test/screens/search_screen_test.dart` — **16/16 passing**.

New/updated tests cover: `submitSearchQuery` forwarding through the mixin, the TV-OSK search key moving focus to the first result, a companion-remote submit dismissing an open OSK and focusing results, and a no-results remote submit focusing the input without the OSK.

## Notes

- Only touches Dart source + tests; no native, dependency, or generated-file changes.
- No user-facing string changes, so no `slang` regeneration needed.

### DIFF
--- a/lib/focus/focusable_text_field.dart
+++ b/lib/focus/focusable_text_field.dart
@@ -26,6 +26,30 @@ enum TvKeyboardAutoOpenBehavior {
   never,
 }
 
+/// Imperative handle to the TV on-screen keyboard of a [FocusableTextField] /
+/// [FocusableTextFormField]. Pass the same instance to the field's
+/// `tvKeyboardController`; the field's host attaches itself on mount.
+///
+/// Only meaningful on TV. On other platforms the OSK never exists, so every
+/// method is an effective no-op.
+class TvKeyboardController {
+  _FocusableTextInputHostState? _host;
+
+  void _attach(_FocusableTextInputHostState host) => _host = host;
+  void _detach(_FocusableTextInputHostState host) {
+    if (identical(_host, host)) _host = null;
+  }
+
+  /// Dismiss the OSK if it is open (and prevent it from auto-reopening while
+  /// the field keeps focus). No-op when nothing is open.
+  void closeKeyboard() => _host?._dismissTvKeyboard();
+
+  /// Focus the field but do NOT open the OSK for this focus entry. Used as a
+  /// fallback landing spot (e.g. a remote search that returned no results) so
+  /// the remote isn't stranded on an off-screen element.
+  void focusInputWithoutKeyboard() => _host?._focusWithoutKeyboard();
+}
+
 String _describeTextInputKey(KeyEvent event) {
   return 'type=${event.runtimeType} logical=${event.logicalKey.keyLabel}/${event.logicalKey.keyId} '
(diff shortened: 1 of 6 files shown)


edde746:
~~AI used? If so, which model?~~

My bad, I see it on the commit.

l3gitpanda:
> ~AI used? If so, which model?~
> 
> My bad, I see it on the commit.

All good. I'm unable to test the fix due to not having a mac and not having a dev account for testing on appletv but im hoping this fixes the issue.

PR #1524, opened 2026-07-10
feat(remote): companion remote playback bar, chapter skip, and Explore search

l3gitpanda:
## Summary

Quality-of-life improvements to the companion remote (phone controlling a desktop/TV Plezy), plus two fixes found while testing them:

- **Persistent playback bar** — while the host player is active, a transport bar (prev chapter · seek back · play/pause · seek forward · next chapter) stays pinned at the bottom of the remote on all three tabs, so seeking no longer requires switching to the Play tab.
- **Chapter skip** — new `previousChapter`/`nextChapter` protocol commands wired to the video controls' existing chapter seek (falls back to a small time seek when the media has no chapters). Buttons live in the persistent bar and on the Play tab.
- **Explore from the remote** — new `tabExplore`/`exploreSearch` commands: an Explore chip in the remote's navigation (opens the search sheet and switches the host to Explore) and an Explore card on the More tab. The typed query opens `CatalogSearchScreen` on the host pre-filled via a new `initialQuery` param, using the same TV on-screen-keyboard suppression as `SearchScreen.submitSearchQuery`.
- **Navigation chips stay visible during playback** (previously hidden while a player was active).
- **Fix: play/pause on the remote did nothing on desktop** — it was a simulated space-bar key press, which only walks the focus tree and never reaches the player's `HardwareKeyboard` shortcut. It now has a real `onPlayPause` callback (key simulation remains as app-level fallback).
- **Fix: remote-driven searches ran invisibly** beneath pushed screens — remote search commands now pop to the main screen first, and repeated Explore searches replace the catalog search screen instead of stacking.

## Protocol compatibility

`RemoteCommandType` is wire-serialized by index (append-only); the four new commands are appended at the end. Older hosts decode unknown indices to the `ping` fallback — a silent no-op, verified against an Apple TV running a current release build.

Note on history: the branch includes a merge + revert pair from an earlier chapter-skip experiment on the fork; they cancel to a net-zero diff.

## Screenshots
image
image
## Testing

- `flutter analyze` clean, `dart format` clean, full `flutter test` suite passing (new widget test covers `CatalogSearchScreen.initialQuery`)
- Manual: iOS phone (sideloaded) controlling a Windows desktop host — playback bar on all tabs, chapter skip with/without chapters, Explore search end-to-end, play/pause, library search after Explore search
- Backward compat: phone with this build against an older Apple TV host — new commands safely no-op

https://claude.ai/code/session_016oGGWGt2igPx8mn8k5dkJb


enchained:
That looks useful, I always felt like companion needs to support more of the features.

Would you consider adding few more playback features to the companion? For example, 2x speed toggle, next audio/next subtitles (or maybe event select from a list), seek buttons for the existing subtitle seek commands, frame by frame seek buttons. It would also make sense to make most extra buttons opt-in via some favorite button menu, so the user can create their personal playback control panel.

l3gitpanda:
> That looks useful, I always felt like companion needs to support more of the features.
> 
> 
> 
> Would you consider adding few more playback features to the companion? For example, 2x speed toggle, next audio/next subtitles (or maybe event select from a list), seek buttons for the existing subtitle seek commands, frame by frame seek buttons. It would also make sense to make most extra buttons opt-in via some favorite button menu, so the user can create their personal playback control panel.

I could look into implementing those features. I also was thinking of adding video player seek functionality to the remote. I'll see what I can do when I find some time.

l3gitpanda:
Closing this one out — upstream's recent remote work covers what we needed, and we're no longer pursuing these changes. Thanks for the earlier review!

PR #1526, opened 2026-07-10
feat(livetv): add guide search with jump-to-channel/program

l3gitpanda:
## What

Adds a search to the Live TV guide: a search icon on the Guide tab opens a sheet that filters **channels** (name, call sign, number) and **programs airing in the next 24 hours** (title / series name) as you type. Selecting a result closes the sheet and jumps to it in the guide — shifting the guide's time window when the airing isn't in the current one, scrolling to the channel row, and landing d-pad focus on the matched program block.

## Screenshots

image

image

image

## How

- New `GuideSearchSheet` modeled on the existing subtitle-language filter sheet (`FocusableTextField` + in-memory filter + `FocusableListTile` rows), shown via `OverlaySheetController.showAdaptive` like the reorder-favorites sheet. Programs come from one `fetchSchedule(now, now+24h)` fan-out over `liveTvServers` when the sheet opens; each program's channel is resolved once so keystroke filtering stays cheap.
- Two new public methods on `GuideTabState` (`jumpToChannel` / `jumpToProgram`) reuse the existing day/time-picker window mechanics, `_scrollToChannel` / `_scrollToProgramTime`, and the `focusContent()` keyboard-mode gating.
- **Companion remote:** while the sheet is open it takes over `CompanionRemoteReceiver.onSearchAction` (same save/restore idiom as the video player's overrides) so a query typed on the phone lands in the guide search with the TV keyboard dismissed; the global Search tab handler is restored on close. No protocol changes.
- Searching covers all channels; jumping to a non-favorite while the favorites filter is on drops the filter so the target row exists.
- 5 new i18n strings (base locale only, `fallback_strategy: base_locale`).

## Testing

- `flutter analyze` clean; `dart format` (120) applied.
- Manually verified on Windows desktop and TV (d-pad): filter/jump for channels and programs incl. window shift for later airings, OSK focus flow, companion-remote query submit into the open sheet, and restore of global search after closing.


l3gitpanda:
Id just like to bump this as I think it is severely needed or an alternative search method for tvOS and other platforms too.

edde746:
Closed in favor of https://github.com/edde746/plezy/commit/a3cbf323b34fd0d7a158361f2753ff97afbae13f

PR #1529, opened 2026-07-11
Add a setting to choose which servers appear on Home

8bitgentleman:
## What this does

When your account is connected to more than one server, the Home (Discover) tab pulls rows from all of them with no way to narrow it down. This adds a "Servers on Home" setting to hide specific servers from Home.

The default is every server, so nothing changes unless you opt in, and the setting only appears when you actually have more than one server connected.

## How it works

The setting stores the *hidden* servers rather than the selected ones (an exclusion set, where empty means show all). Two reasons for that:

- It matches the existing hidden-libraries pattern already in the codebase.
- A server you connect later shows up on Home automatically, instead of being silently absent until you remember to revisit settings.

Filtering happens at fetch time — hidden servers are dropped from the aggregation fan-out, not fetched and thrown away. Both load paths are covered: the full load, and the delta load that runs when a server comes online. That second one matters, otherwise a hidden server coming online mid-session would quietly re-queue itself onto Home.

Toggling applies right away. Hiding a server drops its rows and Continue Watching items in place with no refetch; un-hiding pulls just that server back through the existing delta machinery, no full reload. You can't hide your last remaining server — at least one stays checked.

## Tests

Added coverage in `discover_provider_test.dart` for the cases that actually matter:

- default (nothing hidden) still fans out to every server
- hiding a server drops its content in place
- a hidden server coming online is never delta-fetched
- un-hiding brings a server back via delta, without a full reload
- hiding every server empties Home, and a reload doesn't leak them back
- a stale hidden id (server no longer on the account) loads without error

`flutter analyze` is clean.

## One known rough edge

The "more Continue Watching" indicator can be briefly stale right after you hide a server that contributed items near the preview cap. It fixes itself on the next load. Correcting it immediately would mean a refetch, which didn't feel worth it for a settings toggle — happy to change that if you'd rather it be exact.


edde746:
You can already hide a server by hiding it's libraries.

PR #1574, opened 2026-07-14
fix(player): make desktop chapter selector work

RyanTheTechMan:
Fix issue that prevents mouse users from opening and interacting with the chapter selector on desktop.

## Changes
- Open and close the existing chapter and queue strip using a clickable button
- Scroll the open strip with a mouse wheel or trackpad
- Close the strip when clicking the video

https://github.com/user-attachments/assets/ea801964-4f06-40f0-83cc-1f7517cfd898


edde746:
That chapter selector is only visible on desktop if you've enabled keyboard navigation of the player.

PR #1584, opened 2026-07-16
feat(player): add frame-by-frame navigation

RyanTheTechMan:
Add customizable frame-by-frame navigation for videos (when paused), with comma and period as the default shortcuts.
Similar to how it works with YouTube.

## Changes
- Add Previous Frame and Next Frame actions to Keyboard Shortcuts.
- Step paused media via keybinds (or hold down to step faster)
- Show the frame counts skipped in the existing player indicator

https://github.com/user-attachments/assets/bcf7cf44-727b-444c-a802-3e9943e35f96


mr-brune:
Would be very useful even jellyfin is adding it in 12.0

PR #1593, opened 2026-07-16
Sync with upstream 2.9.1 (557 commits); keep both watchlists

TeeJS:
Brings the fork up to date with `edde746/plezy` — **2.3.0 → 2.9.1**, 557 commits. The fork is now **0 behind upstream**.

## Both watchlists are kept, side by side

They looked like duplicates but they aren't — different features, different backends, disjoint gating:

| | Fork's watchlist | Upstream's watchlist |
|---|---|---|
| Backend | plex.tv cloud (`discover.provider.plex.tv`) | Trakt / MAL / Seerr |
| Gated on | a connected Plex.tv account | a connected tracker |
| Surface | its own nav tab | one row on the Explore tab |

`CatalogSourceId` is `{trakt, mal, seerr}` — **Plex is not a catalog source**. Without a tracker connected, upstream's watchlist doesn't exist for you; dropping the fork's would leave no watchlist at all.

## "Plex Watchlist" rename (2nd commit)

Both were labelled plain "Watchlist". Upstream's is disambiguated by context (its row sits under the Explore source header showing the provider logo/name); the fork's top-level tab had no such framing, so the *bare* label was on the wrong one. The fork's is now explicitly "Plex Watchlist" everywhere. Upstream's `explore.*` strings are untouched.

## Notable conflict calls

- **navigation** — `NavigationTabId` carries both `explore` and `watchlist`. Upstream's new `resolveDefaultTab` also threads `hasPlexAccount`; without it the Watchlist tab could never be a startup section. Enum reorder is safe: `EnumPref` serialises by **name**, not index.
- **discover_screen** — the fork's `_buildTvContent` was upstream's *old* hand-rolled spotlight geometry plus a hero-toggle early return. Upstream replaced that body with `TvSpotlightScaffold`, so only the early return is re-applied.
- **appearance_settings** — upstream gates `showHeroSection` to non-TV; the fork's TV hero toggle *is* the removal of that gate, so it stays removed.
- **action_buttons** — upstream's new bookmark button is the *catalog* watchlist, gated on tracker candidates. Its responsive tiers also fix the row overflow the fork removed its own Plex button to avoid.

## Bugs found and fixed along the way

- **`build.yml` duplicate `if:` keys** — git auto-merged `build-ios`/`build-macos` with *two* `if:` keys (upstream's input gate + the fork's skip). YAML silently keeps the last, so `inputs.build_ios` would have been ignored. Now composed: `${{ inputs.build_ios && github.repository == 'edde746/plezy' }}`.
- **`create-release` would have run and failed** — it previously relied on an *implicit* skip (its needs were skipped). Upstream's `always()` defeats that, so it would run against absent iOS/macOS/Linux artifacts. Fork skip restored.
- **`.gitignore`** — `server/deploy.shbuild-output/` was a missing newline that broke *both* entries. That's why `build-output/` showed up untracked.
- **`Casks/plezy.rb`** — takes upstream's 2.9.1. The fork's `windows-latest` was `update-packages.yml` stamping a rolling tag into a **macOS** cask pointing at upstream releases → a URL that doesn't exist.

## Fork code vs upstream refactors it predates

`getPlexClient` now takes `ServerId`; `_buildNavItem` lost `isFocused`; the side-rail switch needed its `watchlist` case; `_buildTvExpandedContent` moved onto `_discover.*`/`_orderedHubKeys`; `plex_watchlist_service` needed initializing formals for the new analyzer gate. Release actions are pinned to the SHAs upstream already vets. The dead `watchlist.add` string is gone — the "Add to…" dialog title supplies that prefix, so it was never referenced.

## Translation hygiene: deliberate fork divergence

The watchlist strings stay English-only, relying on slang's documented `fallback_strategy: base_locale`. Upstream's new hygiene gate would fill the other 15 locales with `""`, which slang emits **verbatim** — a blank Watchlist label in 15 languages. Verified: `@override String get title => '';`. The fork runs the unused-key scan instead; the full check still runs upstream.

## Verification

Analyzer, codegen, formatting, icon consistency, unused code, unused files, all workflow guards, and **3312 unit tests** pass.

Two local failures are Windows-only and pre-existing in upstream's new tests — `cycling_media_backdrop` (temp-dir file locking in `tearDown`) and `test_check_icon_consistency` (`subprocess` can't resolve `dart` vs `dart.bat` without `PATHEXT`). CI runs both on `ubuntu-latest`.


TeeJS:
Opened unintentionally against the upstream repository by a tooling misconfiguration — this is a fork-internal sync branch and was never meant to be proposed upstream. Closing immediately; apologies for the noise.

PR #1599, opened 2026-07-17
Make grid spacing between posters configurable

### LINKED ISSUE #1597: Make grid spacing between posters configurable
### Problem Statement

The gap between posters in the library grid is currently hardcoded to zero (`crossAxisSpacing = 0` / `mainAxisSpacing = 0` in `lib/utils/layout_constants.dart`) — the only visible spacing comes from the cards' internal padding. The Library Density slider changes poster *size*, but not the breathing room *between* items. Since #629 tightened things up, the grid can feel quite cramped for those of us who prefer a more airy layout.

### Proposed Solution

Add a setting to control the spacing between grid items — e.g. a small slider or a few steps (tight / normal / spacious) next to the existing Library Density option in Appearance settings. It could simply feed into `crossAxisSpacing` / `mainAxisSpacing` of the grid delegate.

### Alternatives Considered

- Cranking Library Density to 5 — makes posters bigger, but doesn't add space between them.
- Tying the spacing to the density value instead of a separate setting would also work, if a second knob feels like too much.

### Use Case

1. As a user who found the pre-#629 spacing more pleasant, I'd bump the spacing up a notch and keep my preferred poster size via density.
2. Others who liked the change in #629 keep the current tight look — everyone's happy with one setting.

### Additional Context

#629 went in the opposite direction (reducing padding). This request isn't about reverting that, just about making it adjustable since taste clearly differs here.

### Target Platform

All platforms

### Media Server

Jellyfin, Plex (UI-level, server-agnostic)

slevin-7:
Closes #1597

Adds a **Grid Spacing** option (Tight / Normal / Spacious) right below Library Density in Appearance settings, feeding `crossAxisSpacing`/`mainAxisSpacing` of the media grids — plus the horizontal shelf rows on home/discover and the TV browse rail.

A few notes on the approach:

- Default is **Tight (0px)** — exactly the current look, so nobody's layout changes on update. Normal is 6px, Spacious 12px.
- The value is read in `MediaGridDelegate.spacingFor()`, which was already the single funnel for non-full-bleed grid spacing, so all grid surfaces (library browse, collections, playlists, downloads, detail screens) pick it up consistently.
- Instead of adding the new pref to every screen's `SettingsBuilder` watch list, it's watched once in `MediaCardSliverLayout` — the only `MediaGridGeometry.resolve()` call site — so grids re-layout live when the setting changes.
- Second commit applies the same setting to the home/discover hub rows: per-card padding grows by the gap, `_itemExtent` (keyboard scroll math) accounts for it, and the card memo epoch includes it.
- Third commit does the same for the TV home screen, which renders shelves through `TvBrowseRail` instead of `HubSection`: the gap feeds `itemGapForScale` (previously hardcoded 0), and card fitting / item extents / d-pad scroll math pick it up through the shared metrics struct. Verified on a Fire TV stick.
- If you'd rather keep the setting scoped to actual grids only, happy to drop commits 2–3.
- The TV full-bleed card layout keeps its own scale-based spacing and is not affected.
- `_calculateInitialFetchSize()` in the browse tab now uses the same spacing for its row-height/column estimate.
- Went with the three steps from the issue rather than a slider — matches the existing segmented-control pattern in that settings group.

Added English + German strings (other locales fall back to English per slang config) and a widget test that flips the setting and checks the grid delegate picks it up.

`flutter analyze` clean, full test suite passes (3339 tests). Manually verified on macOS (library grid + home rows) and Fire TV (browse rail) — all re-layout live when toggling the setting.


albino1:
I wonder if this could be looked at and merged.  Right now the spacing is extremely tight, it would be nice to be able to pad it out a bit.

Issue #1518, opened 2026-07-10
Add Director to Media Info


R3iniX:
### Problem Statement

It would be helpful to quickly look for other movies of specific director.

### Proposed Solution

- Add Director to the Movie Info screen e.g.
-- underneath the Genre bubble (like on Plex Web)
-- As part of the Cast Section
- make it clickable to view other movies of director

### Use Case

View other movies of director.

### Target Platform

All platforms

### Media Server

Plex

### Additional Context

<img width="1047" height="465" alt="Image" src="https://github.com/user-attachments/assets/7899430f-f3da-4c85-91a9-ee1a9a95a75b" />

hivehome:
My daughter like to research directors and used this feature in plex as a learning tool. It shows complete assuming this is coming in update.

PR #1633, opened 2026-07-19
feat: Add Director to Media Info Rows

ejach:
This PR adds the `Director` that is already present in pulled metadata to the info rows. This includes an addition to all of the translation strings available. This handles the case of `Directors.length >= 2` and the case of no directors.

This somewhat addresses #1518 but is not present on TV view ports.
I am currently working on the TV view port and will address this in a separate PR.

**Documentation**

**Sahara (1943)**
Screenshot_1784498925

**No Country for Old Men (2007) -> the Cohen Brothers**
Screenshot_1784499714

### DIFF
--- a/lib/i18n/bg.i18n.json
+++ b/lib/i18n/bg.i18n.json
@@ -716,6 +716,8 @@
     "extras": "Трейлъри и екстри",
     "studio": "Студио",
     "rating": "Рейтинг",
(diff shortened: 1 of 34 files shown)


edde746:
Merged in https://github.com/edde746/plezy/commit/c9543d4af0e6d010ff33d2ca8636d2cdd873f2de

PR #1643, opened 2026-07-20
Add custom authentication header support for Seerr

8KV:
Adds optional custom authentication headers for Seerr, providing an additional security layer and supporting setups such as Cloudflare Access or similar services.


mvonpohle:
This would close #1877

PR #1683, opened 2026-07-25
i18n: Add Turkish translation

Omc725:
Added Turkish language support via tr.i18n.json file.


edde746:
Thanks @Omc725 — this is merged as part of a consolidated Turkish commit.

Your translation was contributed against an older English source, so it was missing 51 keys (most importantly the entire `downloads.backgroundWarning` block, which `test/i18n/locale_fallback_source_test.dart` requires every locale to translate), and it carried an orphaned `downloads.manageSync` key. The JSON on its own also wouldn't build: `_getLanguageDisplayName` in `lib/screens/settings/appearance_settings_screen.dart` is an exhaustive switch over `AppLocale`, and the generated `strings_tr.g.dart` was missing.

The consolidated commit fills those gaps, registers Turkish in the language picker, runs `dart run slang`, and fixes a few strings from the contribution: `common.mute` ("Sesi Kıs" = volume down, colliding with the volumeDown label), `music.discNumber` ("Disket" = floppy diskette), the three "bitiş hızı" bitrate keys ("bitiş" = ending), and a `Kısayol Ayaıla` typo.

PR #1684, opened 2026-07-25
Guard linux is available around actually installed linux commands.

aldobarr:
Currently on linux the app lists all of the linux supported player options even if they're not actually installed.

This adds a small helper to check if a player is actually available before listing it to the user.


edde746:
Landed in https://github.com/edde746/plezy/commit/9cd486f5a2c46b040bae80fc996918860af26929

PR #1689, opened 2026-07-26
i18n: Add Azerbaijani translation

Omc725:


edde746:

A few things had to be completed before it could ship. The translation was contributed against an older English source and was missing 51 keys, including the whole `downloads.backgroundWarning` block that `test/i18n/locale_fallback_source_test.dart` requires every locale to translate. `downloads.syncRuleCreated` had dropped the `${count}` placeholder, which breaks the generated signature. The JSON alone also wouldn't build: `_getLanguageDisplayName` in `lib/screens/settings/appearance_settings_screen.dart` is an exhaustive switch over `AppLocale`, and `strings_az.g.dart` was missing.

The consolidated commit fills the gaps, restores `${count}`, registers the locale, and runs codegen. It also corrects `fileInfo.resolution`, which read "Imkan (Kopyalama)" ("possibility (copying)") instead of Resolution, the non-word "oynatranı" in `settings.bufferSizeWarning`, a leaked "(Kadrı)" note on `liveTv.record`, and a handful of Turkish spellings (gösterilən/göndərin/Nə/Ən/artıq/tarixi/Kitabxana).

PR #1690, opened 2026-07-26
Create kk.i18n.json

Omc725:


edde746:

Same gaps as the sibling PRs: contributed against an older English source, so 51 keys were missing (including the whole `downloads.backgroundWarning` block that every locale must translate), `downloads.syncRuleCreated` had dropped its `${count}` placeholder, and the JSON alone wouldn't build without the `AppLocale` switch case in `lib/screens/settings/appearance_settings_screen.dart` plus generated `strings_kk.g.dart`.

Kazakh also needed some cleanup: `common.later` was Latin-script "Keyinirek" on a common dialog button, "Subtitr" appeared untransliterated in five keys, and a few Uzbek/Tajik forms had leaked in — "бош" (Kazakh: "бос"), "бир" and "Киру" (Kazakh uses і, not и), "муаммо", and the non-word "Аулымен". `watchTogether.title` also read "Бетпе-бет көру" ("face to face"), which mislabels the Watch Together feature; it is now "Бірге көру".

PR #1691, opened 2026-07-26
Create uz.i18n.json

Omc725:


edde746:

Same gaps as the sibling PRs: contributed against an older English source, so 51 keys were missing (including the whole `downloads.backgroundWarning` block that every locale must translate), `downloads.syncRuleCreated` had dropped its `${count}` placeholder, and the JSON alone wouldn't build without the `AppLocale` switch case in `lib/screens/settings/appearance_settings_screen.dart` plus generated `strings_uz.g.dart`.

Your orthography was clean throughout — every `oʻ`/`gʻ` uses U+02BB correctly and the glottal stops use U+02BC — so the additions follow that convention. A few contributed strings are corrected: "tugʻridan" → "toʻgʻridan", "Vaqratlarda" → "Varaqlarda", "ulrang" → "ulang", "davam" → "davom" in five keys, and `settings.autoSkipCredits` "Titr hisoblari" ("credit accounts") → "Titrlar".

Issue #970, opened 2026-05-03
Plex Preroll support


bart7782:
### Problem Statement

Plezy does not play Plex Prerolls, which are short video clips configured in Plex Media Server settings that automatically play before a movie or show starts. This is a native Plex server feature.

### Proposed Solution

Before starting playback, check the Plex server for a configured preroll and play it first, just like the official Plex client does.

### Alternatives Considered

There is no workaround. Prerolls are either played by the client or skipped entirely. Currently Plezy skips them.

### Target Platform

All platforms

### Media Server

Plex

### Additional Context

https://support.plex.tv/articles/202920803-extras/#:~:text=%3A%20Plex%20Subscriptions-,Advanced%20Settings,-Click%20on%20the

jyc66:
FWIW I use pre-rolls and in plezzy it failed to play anything at all. I had to remove the preroll list from my plex server before I could play a movie.

PR #1695, opened 2026-07-28
feat(Preroll): add pre-roll playback before a movie

Bergasha:
This adds the ability to select a number of pre rolls from your library and play a random one before a movie, Press back to skip. Settings options under Video Playback.
Screenshot 2026-07-28 155305


edde746:
Thanks for this — pre-rolls are a genuinely wanted feature, and the settings shape you landed on is a reasonable direction. It needs another pass before it can go in: the branch doesn't currently compile, three CI gates are red, and the playback path reports pre-rolls to the media server as real watches. One triage note up front: this is a Plezy-local picker rather than Plex's native server-side pre-roll, so it doesn't close #970 — that issue should stay open.

## Blocking

### 1. Unrelated revert of the settings-tile refactor → 17 compile errors

`playback_settings_screen.dart` rewrites five unrelated tiles from `SettingSelectionTile<T>` / `SettingSegmentedTile<T>` back to a two-type-parameter form with `decode:`/`encode:`, and deletes two explanatory comments. Current `lib/widgets/setting_tile.dart:176,257` declares **one** type parameter and no codec params.

This isn't stale-base drift — it's in the three-dot diff against your own merge base (`9a0e9611`), so it fails both on the PR head and after merging `main`:

```
error - playback_settings_screen.dart:367:38 - The type 'SettingSegmentedTile' is declared with 1 type parameters, but 2 type arguments were given
error - playback_settings_screen.dart:375:5  - The named parameter 'decode' isn't defined
... 17 errors total
```

`flutter test test/screens/settings/playback_settings_screen_test.dart` fails at load with the same errors. Looks like an editor working from an older copy of the file. Easiest fix: reset `playback_settings_screen.dart` to `main` and re-apply only the pre-roll additions.

### 2. `use_build_context_synchronously` x5 → analyzer gate fails

`scripts/check_analyzer.dart` runs `dart analyze --fatal-infos` against a closed two-entry allowlist, so these are hard failures:

```
info - utils/video_player_navigation.dart:258:16, 260:34, 261:28, 264:19, 265:31
```

The cause is real, not cosmetic. `video_player_navigation.dart:250-256`:

```dart
final preroll = await pickRandomPreroll(context);
if (preroll != null && context.mounted) { ... if (!context.mounted) return null; }
// falls through here with an unchecked context when preroll == null
```

`pickRandomPreroll` awaits `SettingsService.getInstance()` *before* reading the enable flag. The feature is correctly off by default, but this async lookup is not gated by it — so every movie launch, for every user, now crosses an async gap before `context.readFreshWatchState`, `Navigator.of(context)` and four `context.read` calls. Any caller whose context goes away in that gap throws. Please gate the pref synchronously (`SettingsService.instance.read`, as `setting_tile.dart:155` does) so the disabled path stays fully synchronous, and re-check `context.mounted` once after the gap.

### 3. `dart format` gate fails

`scripts/ci_checks.sh:52`. Two files change under the repo's `page_width: 120` (`analysis_options.yaml:84`), and `video_player_navigation.dart:250` is 125 columns.

### 4. Pre-rolls are reported to the server and resumed as real watches

The pre-roll is played by recursing into `navigateToVideoPlayer`, i.e. a full `VideoPlayerScreen`, which has no reporting opt-out (`video_player_screen.dart:275-318` — `live` is the only special mode today). So a pre-roll gets:

- **server timeline/progress reporting and watched-marking** — `playback_progress_tracker.dart` has no `MediaKind` branch, so this is unconditional. The clip lands in Continue Watching and gets marked watched.
- **a saved resume position** — a pre-roll you skip at 3s resumes at 3s next time instead of restarting. This is arguably the most visible symptom.
- **Discord Rich Presence** (`playback_services.dart:648-654`) — not kind-gated.
- **Trakt / tracker scrobbles** *when the chosen items are movies or episodes*. `trakt_scrobble_service.dart:224` and `tracker_coordinator.dart:65` both bail out for anything that isn't movie/episode, so genuine `MediaKind.clip` items are safe — but the picker has no kind filter (see below), and putting bumpers in a Plex *Movies*-type library is a common setup precisely because it gives them posters and metadata. Those items are `MediaKind.movie` and will be scrobbled.

`resolveWatchState: false` only skips *reading* fresh watch state; it suppresses nothing outbound. This needs a real playback mode — threaded the way `LiveTvSessionArgs` already is — that disables progress reporting, watched-marking, resume, scrobbling, RPC and `LocalPlaybackHistory` for the pre-roll leg.

### 5. The hook sits in front of the double-launch guards

The pre-roll block is at line 250; `_videoPlayerNavigationInFlightGuard` is marked around 297 and `VideoPlayerScreenState.isNavigationActive` is checked at 357. A double-tap on Play now resolves two *different* random pre-rolls, so neither guard fires on identity, two player routes stack, and the movie launches twice. Move the hook below the in-flight guard, or key the guard on the movie before resolving the pre-roll.

### 6. External player: races on desktop/iOS, reports progress on Android

With `useExternalPlayer` on (`video_player_navigation.dart:312-350`), the recursive pre-roll call hands off to `ExternalPlayerService.launch`:

- **non-Android** (`external_player_service.dart:115`) — `player.launch(url)` returns as soon as the app is spawned, so the movie is launched externally right behind the pre-roll: two concurrent handoffs.
- **Android** (`:101-113`) sequences correctly because it awaits the intent result, but then `_reportAndroidExternalProgress` reports the pre-roll's progress to the server, and `_launchAndroidNative` passes `metadata.viewOffsetMs` as `startPositionMs`, so the external player resumes the pre-roll mid-clip.

Simplest resolution is to skip pre-rolls entirely when external playback is active.

### 7. Bypasses Plezy's focus and dialog-scope conventions

Two separate problems, both from hand-rolling widgets that have established equivalents.

**a) The new controls bypass the `Focusable*` layer.** To be clear about what this does and doesn't break: `SimpleDialogOption`, `CheckboxListTile` and the bare `TextButton` participate in Flutter's default focus traversal, and `WidgetsApp.defaultShortcuts` maps `LogicalKeyboardKey.select` and `gameButtonA` to `ActivateIntent` on Android (`app.dart:1265-1269`), so remote OK and gamepad A do activate them on Android TV. What's missing is everything Plezy's wrappers add on top: focus chrome gated on input mode (`focusable_wrapper.dart:541` — `showFocus = _isFocused && isKeyboardMode`, so raw Material rows show Material's own highlight unconditionally instead of the app's border/scale/background treatment), `autoScroll` into the viewport (`focusable_wrapper.dart:132`), `onBack` handoff and `BackKeyCoordinator` dedup, and the regression coverage in `test/focus/focusable_wrapper_test.dart` / `test/widgets/focusable_list_tile_test.dart` that the raw widgets don't inherit. Worth double-checking on tvOS specifically, since it resolves to `_defaultAppleOsShortcuts`, which omits both keys — I couldn't verify that on a device.

| Need | Use |
|---|---|
| library picker | `showSelectionDialog<T>` + `DialogOption` (`settings_utils.dart:29,163`) |
| multi-select rows | `FocusableCheckboxListTile` (`focusable_list_tile.dart:308`, see `seerr_request_sheet.dart:488`) |
| dialog action | `DialogActionButton` |

**b) Bare `showDialog` escapes the profile session.** Flutter's `showDialog` defaults to `useRootNavigator: true`, so both pickers mount on the root navigator and outlive a profile switch, which replaces the keyed session subtree they read `LibrariesProvider` from. `showScopedDialog` (`utils/dialogs.dart:18-29`) exists for exactly this — it passes `useRootNavigator: false` so the route is disposed with the session. `lib/` has 3 bare `showDialog<` calls against 28 `showScopedDialog<`.

Unrelated to focus, in the same widget: `SizedBox(height: 400)` is a fixed device assumption, and `ListView(shrinkWrap: true, children: [...])` materialises all 200 rows — `ListView.builder`.

### 8. All strings hard-coded English

Ten-plus user-facing strings (`'Prerolls'`, `'Play prerolls before movies'`, `'Not set'`, `'Pick a library first'`, `'Select prerolls'`, `'Done'`, ...) with no `lib/i18n/en.i18n.json` keys. Convention is nested camelCase keys plus `t.settings.*`; every neighbouring group title is localized.

### 9. "No selection" defaults to "play anything in this library"

The feature flag itself is correctly off by default. But `prerollSelectedItemKeys` defaults to `const []`, and `preroll_service.dart` reads that as *all* items (`selectedKeys.isEmpty ? page.items : filtered`, surfaced as `'All clips in the library'`). Combined with the missing library-kind filter, enabling the toggle and picking a library means a random arbitrary item plays before every movie — a whole feature film if that library holds movies. Please make the empty selection mean "no pre-roll" and require explicit per-clip opt-in.

## Should fix

- **New prefs are in none of the three closed registries** (`settings_service.dart:829-945`). `playPrerollsBeforeMovies` belongs in `_portableOnlyPrefs` alongside `startInFullscreen` / `autoSkipIntro`; the server-scoped `prerollLibraryGlobalKey` / `prerollSelectedItemKeys` belong in `_resetOnlyPrefs`. As written they survive "Reset All Settings" and are dropped by settings export.
- **Server/library-scoped keys stored in global prefs.** `prerollLibraryGlobalKey` embeds a `ServerId` and the item keys are per-account, but `LibrariesProvider` lives in the profile-scoped subtree (`profile_session_screen.dart:227`). After a profile switch the key resolves to nothing and the feature silently dies — or resolves a library the new profile shouldn't see. Needs profile scoping the way `HiddenLibrariesProvider` does (`hidden_libraries_provider.dart:41-52`).
- **No library-kind filter.** The picker lists every library from `LibrariesProvider.libraries`, including music, photo and show libraries. Point it at a show library and `page.items` are shows, which `VideoPlayerScreen` can't play. Filter on `MediaLibrary.kind`. (Listing *hidden* libraries is fine and probably desirable — that's where a pre-roll library would live.)
- **`limit: 200` is a silent correctness hole, and the wrong API.** Selections are stored as `globalKey` but resolved at playback by re-listing the library and filtering, so anything outside the first 200 items in default sort order can never be selected; if none of the selected keys land in that window, `pickRandomPreroll` returns `null` and the feature does nothing with no diagnostic. Two better paths already exist: explicit selection → `client.fetchItem(id)` (`media_server_client.dart:167`); "all clips" → `LibraryQuery(limit: 1, sort: LibrarySort(field: 'random'))` (`library_query.dart:14`). Either one also removes a full library listing from the critical path of every movie launch.
- **Two sequential player sessions.** Pre-roll route → pop → movie route means two full mpv/ExoPlayer inits, and on desktop `exitFullscreenOnPlayerClose` fires between them, so the window leaves and re-enters fullscreen mid-sequence. Playing the pre-roll inside the movie's session (as a pre-playback item) avoids that; if two routes stay, at minimum suppress the fullscreen exit for the pre-roll leg.
- **`onTap: library == null ? () {} : ...`** — a focusable row that swallows Select with no feedback. `SettingNavigationTile` asserts one of `destinationBuilder`/`onTap`, so either add a disabled state or hide the row until a library is chosen. Relatedly, the two dependent rows show even when the toggle is off; neighbouring groups hoist that visibility into the `SettingsBuilder` in `build()` — which is what the comment this PR deleted was explaining.
- **`_error = e.toString()` rendered in the dialog.** Raw exception text can carry server URLs, and `getMediaClientForLibrary` throws a bare `Exception`, so the user sees `Exception: No client available`. Use `appLogger.w(..., error:, stackTrace:)` plus a localized message through the shared snackbar helpers.
- **File placement / layering.** `preroll_service.dart` is a `*_service` living in `lib/utils/`, and it takes a `BuildContext` to do `context.read<LibrariesProvider>()` plus HTTP. That inverts the documented flow (`screens → providers → media contracts → services`). Resolution belongs in `lib/services/` behind injected dependencies, with `BuildContext` kept at the call site.
- **`svc.write(...)` unawaited** in the checkbox `onChanged` — wrap in `unawaited(...)`; rapid toggles currently race.
- **No tests.** At minimum: pre-roll skipped when the pref is off, for episodes, for in-progress movies, and offline; a selected key that no longer exists degrades to no pre-roll instead of throwing; and a focus test for the pickers wrapped in `InputModeTracker`.

## Nits

- `Random()` per call is fine, but a "don't repeat the last clip" rule would noticeably help a 2–3 clip library.
- Two `Builder`s each calling `context.watch<LibrariesProvider>()` rebuild the whole group on every library refresh; one `Selector` would do.
- `usePushReplacement` silently disables pre-rolls; today that's only `episode_navigation.dart:121` (episodes, never movies), so the condition is dead weight — drop it or comment why it's there.
- `hasActiveProgress` gating means "Play from beginning" on a partially-watched movie gets no pre-roll. Probably worth wiring to the explicit-restart path.

## What I ran

| Check | Result |
|---|---|
| `dart analyze` on PR head (`ca9a081c`) | 22 issues — 17 errors, 5 fatal infos |
| `dart analyze` after merging `main` (`a921edd2`) | 22 issues, identical |
| `dart format --set-exit-if-changed` (3 touched files) | 2 files changed → fail |
| `flutter test test/screens/settings/playback_settings_screen_test.dart` | fails at load (compile errors) |
| runtime smoke test | not possible — branch doesn't compile |

Happy to look again once it builds; the settings surface is close to right, and items 4/5/6 are really one change (a non-reporting pre-roll playback mode) rather than six.


---

*Edit: corrected blocker 7a. I originally claimed remote OK / gamepad A wouldn't activate the raw Material controls at all — that's wrong; Flutter's default shortcuts map both to `ActivateIntent` on Android (`app.dart:1265-1269`). Rescoped to the conventions actually bypassed. Everything else stands.*

Bergasha:
That is one detailed reply haha, Thanks for that, I'll jump in and get working on this over the next few days.

Bergasha:
@edde746  All tests have now passed. This should be ready for a re-review. Thanks again for the detailed reply, it was really helpful.
I wonder if this should close #970 as Plex has no API for their existing pre-rolls so this will be as good as we get for the time being?.

Bergasha:
@edde746 When you get a second, this is ready for a review. Thanks mate

wraslore-hub:
Did this feature get put in? I don't see anywhere you can set pre rolls in plezy.

Bergasha:
> Did this feature get put in? I don't see anywhere you can set pre rolls in plezy. 

I just pushed through another PR for this yesterday.

PR #1737, opened 2026-07-31
Add Emby and Jellyfin client implementations with UI updates

Omc725:


edde746:
AI used? If so, which model?

Omc725:
Sonnet 4,6 and gemini 3,6 flash

PR #1764, opened 2026-08-01
feat: add local time digital clock to top app bar

dhanict:
Hi!
I've added a simple digital clock to the top app bar that displays the user's local time.

Why?
When using the app on a TV, it's very easy to lose track of time. Having a small, non-intrusive clock at the top right helps users keep an eye on the time while navigating.

It matches the font weight, margins, and styling of the surrounding action icons. Let me know what you think or if you need me to adjust anything! 

image


micahmo:
Nice, Plex for TV has this too. Does the time format respect the user's locale for 12/24 hr?

Bergasha:
> Nice, Plex for TV has this too. Does the time format respect the user's locale for 12/24 hr?

Good question; hopefully it can be set to just AM-PM like this.
Screenshot 2026-08-05 075309

PR #1786, opened 2026-08-04
fix(linux): do not schedule audio recovery after playback resume

torincb:
This was resulting in two audio stutters per playback resume.

Regression originates in 9f2e05079 (release 2.10.0).


edde746:
This seems to have unrelated changes

torincb:
@edde746 Sorry, I don't understand - this change is very targeted. I removed a block which causes an audio recovery sequence upon normal playback resume, which is unnecessary and results in stutter.

Can you let me know what you'd like to see different? I will raise a corresponding issue if you like - the problem is very well scoped.

edde746:
Not sure what happened, maybe I was looking at the wrong thing.

torincb:
Thanks, appreciate you acting fast 😁 

Just want to take the opportunity to thank you for this app - major step up from the competing clients in pretty much every way.

PR #1793, opened 2026-08-04
feat(androidtv): caption service

watermarkhu:
Listen to [`KeyEvent.CAPTIONS`](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_CAPTIONS) on Android TV such that the captions remote button triggers the subtitle selector. 

Implemented with opencode


edde746:
Thanks for this — the behavior is the right idea, but I think the whole native layer can go, and there's a latent lifecycle bug in the dispose guard. Three blocking items plus a UX question.

### 1. The method channel is redundant — Flutter already delivers this key

Flutter's Android embedder maps `KEYCODE_CAPTIONS` (175 / `0xaf`) to `LogicalKeyboardKey.closedCaptionToggle`, in the same table as the media keys we already handle in Dart:

```
// io/flutter/embedding/android/KeyboardMap.java — Android keyCode → logical key
put(0x00000000afL, 0x0100000d12L); // closedCaptionToggle
put(0x000000005aL, 0x0100000d2cL); // mediaFastForward
```

Verified on an NVIDIA SHIELD (`ro.build.characteristics=tv`) with a stock Flutter app — no native code, no channel — using `adb shell input keyevent 175`, so it exercises the keycode path rather than one remote's particular button:

```
I flutter : KEYPROBE KeyDownEvent logical=Closed Caption Toggle physical=0x11000000af isCC=true
I flutter : KEYPROBE KeyUpEvent   logical=Closed Caption Toggle physical=0x11000000af isCC=true
```

`key_events.dart:242-257` already consumes `mediaFastForward` / `mediaTrackNext` from TV remotes this way. Also worth noting the `return true` in `dispatchKeyEvent` consumes the key *before* `FlutterView`, so the native hook blocks the framework path rather than filling a gap in it.

I tried the Dart-only version locally: reverted `MainActivity.kt` and `main.dart`, deleted `captions_service.dart` and its test plus the `initState`/`dispose` wiring, and added this to `_handleGlobalKeyEvent` (focus-independent, placed before the sheet-open bail so a second press still closes the sheet):

```dart
// The subtitles/CC key on TV remotes (KEYCODE_CAPTIONS) is focus-independent
// and toggles the track sheet, so it is handled before the text-input bail
// below: it can never type, and the sheet holds focus while it is open.
if (event is KeyDownEvent && event.logicalKey == LogicalKeyboardKey.closedCaptionToggle) {
  _toggleSubtitleSelector();
  return true;
}
```

`track_controls.dart` stays exactly as you wrote it. Your three test cases still pass with the widget test retargeted to `sendKeyDownEvent(LogicalKeyboardKey.closedCaptionToggle, physicalKey: PhysicalKeyboardKey.closedCaptionToggle, platform: 'android')`. That's +8 lines in one file in place of 50 lines across four files and two languages, plus the 38-line service test. It also picks up the `ModalRoute.isCurrent` guard that `_handleGlobalKeyEvent` already applies; the native path has no equivalent, so by inspection CC would open a sheet even with a route above the player (I didn't test that scenario). Happy to hand you the patch.

### 2. The dispose guard never fires, so a disposed State stays reachable

Mount the controls, unmount them, and `CaptionsService.onToggleSubtitleSelector` is still non-null — confirmed with a probe widget test on this branch. The guard

```dart
if (identical(CaptionsService.onToggleSubtitleSelector, _toggleSubtitleSelector)) { ... }
```

compares against a freshly torn-off extension method (`_toggleSubtitleSelector` lives in `extension _PlexVideoControlsTrackMethods`), which isn't a stable identity token: two tear-offs of the same extension method on the same receiver compare unequal under both `identical` and `==`.

So the field keeps a closure over the disposed `_PlexVideoControlsState` (→ widget → player, chrome controller, metadata) until something replaces it — a later `PlexVideoControls` mount overwriting the field, an explicit assignment to `null`, or process exit. That's one stale State at a time rather than an unbounded leak, but a disposed State stays reachable after the viewer leaves the player, and the "Cleared on dispose" doc comment doesn't hold. The current tests can't catch it: the group's `tearDown` clears the static itself, and `if (!mounted) return` hides the behavioral symptom.

`screens/video_player/parts/pip.dart:15-19` with `playback_start.dart:320-321` shows the working pattern — keep the closure in a field and compare against that instance. Going the Dart route in #1 removes the need for the callback entirely.

### 3. `dart format` gate

```
Changed lib/widgets/video_controls/parts/track_controls.dart
Changed test/screens/video_player/captions_key_toggle_test.dart
```

The `.show(...)` / `.whenComplete(() {` chain in `_toggleSubtitleSelector` reflows. `flutter analyze` on the touched files is clean.

### 4. The CC button lands D-pad focus on the audio list

`TrackSheet` puts audio in the first `Expanded` and subtitles in the second (`track_sheet.dart:110-118`), and in keyboard/D-pad mode the host focuses `traversalDescendants.first` (`overlay_sheet.dart:451-494`). With a fake exposing two audio tracks and one subtitle track, traversal order is `English, Japanese, Off, English` — audio column first — so the first focused row on a remote is an audio track. When the item has no subtitle controls at all, the sheet is even titled "Audio" (`track_sheet.dart:54-56`).

Either pass an `initialFocusNode` for the subtitle column, or bind the key to `_toggleSubtitles()` (visibility on/off), which is closer to what `KEYCODE_CAPTIONS` means on the platform and needs no sheet at all. Your fake player has one subtitle track and no audio, so the both-columns case is currently untested.

### Minor

- `main.dart:893-895`: two consecutive `if (Platform.isAndroid)` inside `bestEffort('Native window', ...)` — captions isn't a native-window concern.
- `captions_service.dart` skips PiP's `_isAvailable` guard before `setMethodCallHandler`, and the doc comment conflates unconditional *registration* with construction (which is Android-only).
- `captions_service_test.dart`: "resolves the singleton so the constructor registers the channel handler" only asserts singleton identity; it never touches the channel.
- Stray blank line at `video_controls.dart:822`.
- Test lives in `test/screens/video_player/` while the widget under test is `lib/widgets/video_controls/`; and `TvDetectionService.debugSetAppleTVOverride(true)` is a confusing way to fake an Android TV.
- Second press closes whatever sheet happens to be open (settings, chapters, queue) rather than switching to tracks — flagging in case that wasn't deliberate.

Issue #1520, opened 2026-07-10
Pausing a show for a while and resuming causes the playing next to happen in the middle of the episode.


TheDogg:
### Bug Description

After pausing (for a few mins) and resuming the playing next triggers around the time it was originally triggering if the episode was not paused.  Seems to play whatever is left in the buffer and the do the playing next. 

### Steps to Reproduce

Start an episode
pause the episode for 10 mins
resume the episode
playing next will trigger before the end of the episode and automatically skip to the next episode even if you didnt finish watching the current episode

### Expected Behavior

playing next only triggers when the episode is actually over

### Actual Behavior

playing next triggers during the episode

### Plezy Version

2.8.0

### Platform

Windows

### Media Server

Plex

### Device Information

Windows Desktop

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

edde746:
Could you get a log?

TheDogg:
```
[09:48:43.284] [WARNING] [MPV:ffmpeg] https: HTTP error 503 Service Unavailable
[09:48:43.139] [WARNING] [MPV:ffmpeg] https: Will reconnect at 633339904 in 0 second(s), error=I/O error.
[09:48:43.139] [ERROR] [Player LOG ERROR] [ffmpeg] https: Stream ends prematurely at 633339904, should be 2345129987

[09:48:43.139] [ERROR] [MPV:ffmpeg] https: Stream ends prematurely at 633339904, should be 2345129987
[09:48:43.139] [ERROR] [Player LOG ERROR] [ffmpeg] tls: IO error: End of file

[09:48:43.139] [ERROR] [MPV:ffmpeg] tls: IO error: End of file
```

PR #1809, opened 2026-08-05
fix(player): hold the Watch Together anchor while the host reloads

toluLikesToCode:
An in-place source switch — audio, subtitle, version or quality — detaches the host's player for the duration of the reload. `_broadcast` falls back to a position of 0 when no player is attached, so any state published in that window names 0:00 as the authoritative position and every guest hard-seeks to the start of the item.

Heartbeats already suppress themselves while detached (`_onHeartbeat` bails on a null player at 738), which is why this hides. The paths that leak the zero are the ones that answer on demand: `onStateRequested`, `onPeerJoined` and `onReconnected` all broadcast regardless of whether a player is attached. A guest entering the player, joining, or reconnecting while the host is mid-switch is the trigger.

Probe against unmodified main — attach a host at 30:00, detach the player, then read `anchorPositionMs` off what gets published:

```
attached=1800000ms onStateRequested=0ms onPeerJoined=0ms onReconnected=0ms
```

Falls back to the last broadcast anchor instead. `_lastBroadcast` is only assigned for untargeted broadcasts, so it holds the last position the room was actually told rather than a targeted reply, and the reload's own re-attach path already re-anchors from it once the player returns. `_pendingStartAtMs` is cleared by `detachPlayer`, so the other branch of `_broadcast` is unreachable in this state and is left alone.

Test asserts all three call sites hold the anchor; it fails on main with `Expected: <120000> Actual: <0>`.

Supersedes the barrier and offset-transcode work that was previously on this branch — see discussion below.

### DIFF
--- a/lib/watch_together/services/host_playback_coordinator.dart
+++ b/lib/watch_together/services/host_playback_coordinator.dart
@@ -777,7 +777,11 @@ class HostPlaybackCoordinator {
       anchorPositionMs = _pendingStartPositionMs ?? player?.position.inMilliseconds ?? 0;
       anchorHostTimeMs = _pendingStartAtMs!;
     } else {
-      anchorPositionMs = anchorPositionOverrideMs ?? player?.position.inMilliseconds ?? 0;
+      // An in-place reload detaches the player, and the reply-on-demand paths
+      // still answer while it is gone. Without the last broadcast to fall back
+      // on, they publish an authoritative 0 and every guest hard-seeks to 0:00.
+      anchorPositionMs =
+          anchorPositionOverrideMs ?? player?.position.inMilliseconds ?? _lastBroadcast?.anchorPositionMs ?? 0;
       anchorHostTimeMs = _nowMs();
     }
 
--- a/test/watch_together/host_playback_coordinator_test.dart
+++ b/test/watch_together/host_playback_coordinator_test.dart
@@ -716,6 +716,36 @@ void main() {
         h.dispose();
       });
     });
+
+    test('the reply-on-demand paths hold the anchor while the player is detached', () {
+      fakeAsync((async) {
+        final h = _Harness(async);
+        h.attachForMedia(async, hasFirstFrame: true);
+        h.hostBecomesReady(async);
+        final anchored = h.last.anchorPositionMs;
+        expect(anchored, const Duration(minutes: 2).inMilliseconds);
+
+        // The state every in-place source reload passes through. Heartbeats
+        // suppress themselves here; the on-demand replies do not, so they are
+        // the paths that would otherwise publish a position of 0.
+        h.coordinator.detachPlayer();
+        async.flushMicrotasks();
+
+        h.coordinator.onStateRequested('guest');
+        h.coordinator.onPeerJoined('guest2', compatible: true);
+        h.coordinator.onReconnected();
+        async.flushMicrotasks();
+
+        for (final (state, toPeerId) in h.sent.skip(h.sent.length - 3)) {
+          expect(
+            state.anchorPositionMs,
+            anchored,
+            reason: 'reply to ${toPeerId ?? 'the room'} must hold the anchor, not collapse it to 0',
+          );
+        }
+        h.dispose();
+      });
+    });
   });
 
   group('driver distraction', () {


edde746 (changes requested):
Thanks for the detailed writeup — the symptom you describe is real, but I don't think this diff is the fix for it, and there are two concrete defects I'd want resolved regardless.

**The premise doesn't match main.** `grep -r 'sourceChangeBarrier' lib/` returns nothing: there is no existing source-change barrier, and the guard that does exist isn't per-switch-kind. `_reloadMediaInPlace` detaches Watch Together at `episode_navigation.dart:625` and re-attaches in `finally` with `hasFirstFrame: _hasFirstFrame.value` (line 969, reset to `false` at line 621), so `HostPlaybackCoordinator.attach()` takes the `sameEpoch && !hasFirstFrame && _phase == playing` branch at line 176 — freeze at the held anchor, resume on all-ready. Audio, subtitle, version and quality switches all reach it.

Offset transcodes likewise aren't unaccounted-for on main; they don't exist. `_buildTranscodeParams` never sends `offset=`, and the current test asserts `expect(startPath, isNot(contains('offset=')))`. The source-local clock that `PlayerBase` now maps around is introduced by this PR.

Size split (`lib/` only, +825/-83): `lib/watch_together/**` is +386/-14, and the offset-transcode machinery (`lib/mpv/**`, `seeking.dart`, `playback_open.dart`, `plex_client.dart`, `sync_offset_control.dart`) is +313/-44. That second group changes every Plex transcode for every user, room or not.

**One candidate for what you saw:** `_broadcast` falls back to `player?.position ?? 0` when the player is null. Heartbeats are suppressed while detached (`_onHeartbeat`, line 738), but `onStateRequested` (line 352) and `onPeerJoined` (line 254) still reply — so a guest entering the player or reconnecting mid-reload receives an authoritative anchor of **0** and snaps to 0:00. Your one-line fallback covers that. Before calling it the fix, though, could you post the log or repro steps? The "host audio change during transcoding" note in `source_switch_barrier_test.dart` suggests you saw something specific, and I'd rather match the fix to it than infer it.

Two defects to fix regardless of how we scope the rest:

**1. `stream-lavf-o` is malformed.** The comment removed at lines 563-566 warns that a multi-code list needs mpv's `%len%` quoting to survive the comma-separated option string — and then `reconnect_on_http_error=404,503` goes in raw. mpv splits `stream-lavf-o` on commas, so `503` lands as a valueless entry. At best the #1520 503 reconnect stops applying to transcode streams; an option parse failure would take `reconnect=1` and `reconnect_delay_max=600` with it. Needs `reconnect_on_http_error=%7%404,503`. As written, the 404 tolerance the readiness preflight is paired with isn't in effect.

**2. `waitForTranscodeReady` floods instead of backing off.** `_http.get` returns non-2xx without throwing — only `throwIfHttpError` throws, and it isn't called here — so a 404, the expected not-ready signal, skips the `catch`, falls through to `consecutiveFailures = 0`, and re-enters the loop with **no delay**. Separately, in the `catch` branch `backoff` is computed, logged, and discarded; the `await` uses the fixed `pollInterval`. Measured on this branch (`pollInterval: 500ms`, `timeout: 2s`, segment always 404):

```
ready=false segmentRequests=12927 elapsed=2001ms
```

Production uses a 15 s window. The new tests miss this: the passing one uses `pollInterval: Duration.zero`, the failing one a 10 ms timeout.

**What I'd like instead:**

1. The `_broadcast` anchor fallback plus a `host_playback_coordinator_test.dart` case asserting that `onStateRequested` with a detached player replies with the held anchor rather than 0 — mergeable on its own once we've matched it against your repro.
2. Whether a *guest's* source switch should freeze the room, as its own change: `_gatingPeers` deliberately doesn't gate a mid-session not-ready guest (line 550, with a comment saying so), and this reverses that. I have further notes on the barrier and protocol-v4 half, but they're only worth working through once we've settled that question.
3. Plex offset transcoding as a separate PR, with the two defects above fixed and the offset applied on initial open too — otherwise seek behavior within one session depends on whether you switched tracks earlier.

Branch also has conflicts in `watch_together_provider.dart` and `host_playback_coordinator.dart`.

toluLikesToCode:
Checked all three against main before replying; you're right on each.

**The premise.** `SourceChangeBarrier` is 0 occurrences on main, 15 on this branch — introduced here, not extended. The guard that exists is the one you name: `episode_navigation.dart:621` sets `_hasFirstFrame = false`, `:626` detaches, `:969` re-attaches with it still false, so `attach()` takes the `sameEpoch && !hasFirstFrame && _phase == playing` branch for all four switch kinds. `offset=` likewise — `_buildTranscodeParams` never sends it and the existing test asserts it doesn't, so the source-local clock this PR maps around is created by this PR.

**The log doesn't support the PR.** I went back to the capture behind the "host audio change during transcoding" note. Its timeline contains `Host source-change barrier anchored at 27986ms` and `TRANSCODE-REQ … offset=27.986000` — both this branch's code. The drift measured there (peak 136.6s, host 4327s vs guest 4190s, sustained 29–34s for ~40s, `ClockSync discarded samples: 0` ruling out jitter) was captured on a build that already carried the barrier, and the repro I wrote down for it starts with `Could not honor source-change barrier` — also this branch's log line. It documents the barrier contending with itself, not a defect on main.

**The anchor-0 defect is real and reproduces on main with none of this branch applied.** Probe against `9d51a040`: attach a host at 30:00, detach the player — the state every in-place reload passes through — then read `anchorPositionMs` off what gets published.

```
attached=1800000ms onStateRequested=0ms onPeerJoined=0ms onReconnected=0ms
```

**Third call site.** You named `onStateRequested` and `onPeerJoined`; `onReconnected` (358) has the same shape and the same result. `_onHeartbeat` bails on a null player at 738, which is why this hides — only the reply-on-demand paths leak the zero.

**Guest source switches freezing the room.** The `_gatingPeers` reversal was incidental, not a position I was arguing for. Dropping it.

**Offset transcoding.** Parking it rather than splitting it out. Main doesn't do offset transcodes at all, so it's a feature rather than a sync fix, and it changes the transcode path for every Plex user whether or not they're in a room. The three defects go with it — including the readiness probe, where you're right that `_http.get` doesn't throw on non-2xx, so a 404 skips the `catch`, resets the counter, and re-enters with no delay.

Reducing this to the anchor fallback across all three call sites plus the `onStateRequested` assertion, rebased onto main.

toluLikesToCode:
I've also deferred the offset-transcode work, to a separate PR that will come later. Need some time to worth through it given the new information

PR #1817, opened 2026-08-06
feat(player): start Plex transcodes at the resume position

toluLikesToCode:
A Plex transcode session always starts producing at zero: `_buildTranscodeParams` never sends `offset=`, so resuming a transcoded title opens a session whose produced window begins at the start of the file and seeks it — the first thing the player does is request a segment the transcoder hasn't reached. This sends `offset=<seconds>` (6dp) with the decision and start request so the session begins producing at the position the player will actually consume: the view offset on initial open, the resolved resume position on every in-place reload.

Nothing about playback timing changes. Measured against a real PMS (movie of 11,845s): the media playlist of an offset session still covers the full title from segment zero (11,845 × 1s segments, `00000.ts`–`11844.ts`), identical to a no-offset session — `offset=` only pre-positions the transcoder. The playlist timeline is the media timeline, so the player keeps opening with `start:` at the resume position and in-stream seeks keep working exactly as today. An earlier draft of this work assumed an offset playlist runs on a source-local clock and built a timeline mapping around that; the measurement above falsified it and the mapping is gone.

Before a native player opens an offset playlist, `waitForTranscodeReady` walks master playlist → media playlist → the segment *containing the offset* (`selectReadinessProbeTarget`, an `#EXTINF` walk), because Plex can publish a manifest before a segment is fetchable and mpv treats that 404 as an HLS error and races through the rest of the manifest. Targeting the offset's segment matters twice over: requesting a segment is how a client seeks a Plex HLS session, so probing the *first* entry would relocate the transcoder back to zero — and probing the right one confirms the server is producing at the target before mpv attaches (observed: resume at 8,062s answered `08062.ts` → 206 on the first probe). The probe treats non-2xx responses as the expected not-ready signal (`_http.get` does not throw on them), waits the 500ms poll interval on every not-ready round, doubles to a 2s cap after three consecutive failed round-trips, and bounds each request with its own timeout so one hung request cannot consume the 15s window. Its regression test asserts a request count, not just the boolean: an always-404 600ms window must stay ≤ 8 requests (the busy-loop variant measured 3,651; the fixed-interval variant 12).

The probe runs only for offset URLs, so no existing open path gains latency; `stream-lavf-o` is untouched (still `reconnect_on_http_error=503` alone — the startup 404 the earlier draft retried is exactly what the preflight absorbs, and an unbounded 404 reconnect would let a dead session spin under a buffering spinner where the #1520 recovery never fires).

A late stop report from the old source can terminate a replacement transcode that shares its session identifier, so in-place reloads now resolve the replacement source only after the stop has gone out.

The old `expect(startPath, isNot(contains('offset=')))` assertion now guards the no-offset case only; a parallel case asserts `offset=495.000000` round-trips decision → start path.

Verified on macOS against a live PMS: resume at 8,062s starts with content at the resume point (previously the transcoder caught up from zero), seeks are unchanged, zero segment 404s in the session log.


edde746:
Reviewed `abd62141` against base `f63d0fe4`.

The core idea is right and the measurement work behind it is unusually good — the falsified timeline-mapping draft in particular. The `offset=` plumbing itself is correct. What I'd block on is the readiness probe: it's a best-effort preflight wired into the shared HTTP client without the guards that surface uses everywhere else, and I measured two side effects plus one behavior regression.

## Blocking

### 1. The probe drives the endpoint-failover cascade

`plex_client.dart:2698` calls `_http.get(...)` with the default `allowEndpointFailover: true`. `failover_http_client.dart:30-32` and `media_server_retry.dart:11-14` both state the rule: a surface with its own retry/timeout budget passes `allowEndpointFailover: false`. This probe has exactly that.

Measured with a scratch test (two prioritized endpoints, probe answers 503):

```
GET …/start.m3u8?offset=10.000000 → 503
Switching Plex endpoint after GET failure
Applying Plex endpoint switch
GET …/start.m3u8?offset=10.000000 → 503     ← same host: the probe URL is absolute
Failover endpoint list reset to first candidate
Applying Plex endpoint switch
REQUEST HOSTS: [primary, primary, primary, primary, primary, primary]   (3 poll rounds → 6 requests)
```

Three consequences:

- The retry is pure waste. `_resolveUri` keeps absolute URLs, so the "failover" re-hits the same host.
- Each switch runs `_handleEndpointSwitch` → `config = config.copyWith(baseUrl:)` plus `_onEndpointChanged` persistence, while `result.videoUrl` already handed to the open path was built from the *old* base.
- On a single-endpoint server `_failoverOnce` takes the `!hasFallback` branch → `onAllEndpointsExhausted`, documented at `failover_http_client.dart:39-41` as the manager's cue to "flip server status and reconnect". That fired three times in a 250 ms window in my run.

Fix is one argument.

### 2. Cancellation is counted as not-ready

`plex_client.dart:2732-2736` catches everything as a transport failure. When the owning client closes mid-probe, `_perform` throws `MediaServerHttpException(cancelled)` before sending, and the loop sleeps out the rest of the window:

```
AFTER CLOSE: ready=false requestsBeforeClose=3 totalRequests=3 blockedForMsAfterClose=2846
```

2.85 s of a 3 s window with zero requests issued — up to 15 s at the production default, all on the playback-open path. AGENTS.md: *"Cancellation is not an empty successful response."* Break on `e is MediaServerHttpException && e.isCancellation`.

### 3. The preflight erases terminal HTTP status

`waitForTranscodeReady` (`plex_client.dart:2703-2739`) treats every non-2xx identically: increment `consecutiveFailures`, sleep, retry until the window closes. `lastStatus` is recorded but only ever reaches a log line at 2748. Callers then map the bare `false` to a single generic error (`playback_start.dart:164-167`, `episode_navigation.dart:708-711`).

The 500 case is the clear regression. `playback_failure_action.dart:46-47` maps 500 → `serverLimitDialog`, shown by `errors.dart:32-35`, with owner-facing copy: *"Server error (HTTP 500). A bandwidth/transcoding limit likely rejected this session. Ask the owner to adjust it."* A preflight failure instead throws `PlaybackException`, caught at `playback_start.dart:414-422`, producing a snackbar reading *"The media server is unavailable. Try again later."* — wrong cause, no remedy, no dialog. `resolvePlaybackFailureAction` never runs at all, because it is fed by `_fatalHttpStatuses` (`video_player_screen.dart:1988`), populated from the **player's** log stream (`errors.dart:60-61`) — and mpv never opens the URL.

This contradicts an invariant the codebase states explicitly, `playback_open.dart:563-565`:

> Deliberately 503 only: a persistent 500 must keep failing fast so the server-limit dialog (`_httpStatusPattern`) appears promptly

The preflight delays exactly that 500 by 15 s and then discards it. Worth noting the PR description makes the same argument against an unbounded 404 reconnect in `stream-lavf-o` — "an unbounded 404 reconnect would let a dead session spin under a buffering spinner where the #1520 recovery never fires" — and the probe reintroduces that shape in Dart for 500, capped at 15 s.

**Fix: the probe should never fail an open — it should only stop waiting.** Drop both throw sites and proceed to open in every non-ready case. mpv then sees whatever the server is actually doing, and the existing log-stream path classifies it exactly as it does today: 500 → server-limit dialog, non-live 404 → media-unreadable dialog. That restores the probe to what it is — a latency and race-avoidance optimization — instead of a new failure mode, and it needs no status plumbed into the UI.

Two adjustments inside the loop follow from that:

- **Exit immediately on 500** rather than polling it for 15 s, so the server-limit dialog appears promptly as `playback_open.dart:563-565` requires. Stop waiting and hand off; don't classify.
- **Keep polling 404 at every depth.** I'm not claiming a playlist-level 404 is terminal — I have no evidence, and neither does the PR: both probe tests 404 only the segment while `start.m3u8` answers 200 (`plex_playback_data_request_test.dart:717-779`, `781-807`), so manifest-404 is unexercised and its meaning unmeasured. Retrying is the right default given the probe exists precisely because Plex's startup window is racy. A 404 that survives the whole window then reaches mpv, which raises the media-unreadable dialog on its own. If you have a PMS trace showing manifest-404 is terminal, exit early on it too and record that in the doc comment.

Not throwing also removes a sharp edge on the reload path specifically: the stop report has already gone out by the time the probe runs, so a thrown exception lands in the rollback at `episode_navigation.dart:947-962`, which resumes a session the server was just told to stop.

## Should fix

### 4. The gate is duplicated verbatim

`playback_start.dart:156-170` and `episode_navigation.dart:701-714` are the same 13 lines. `waitForTranscodeReady` already early-returns for non-`start.m3u8`; fold the `transcodeStreamOffsetFromUrl(...) != null` check in there too — that also closes the footgun where the public method silently probes segment 0, the exact thing its own doc says relocates the transcoder — and share one helper. `_playbackOpenTiming`, `_prepareFrameRateForOpen` and `_resolveOpenResumePosition` are all shared between these two flows for precisely this reason.

### 5. `resolveReplacementAfterStoppedPlayback` earns nothing

`episode_navigation.dart:6-13` is `await x; return f();` behind a public top-level symbol in a `part` file, and its test (`player_initialization_lifecycle_test.dart:48-66`) only proves that `await` defers. Per AGENTS.md a test must "fail on a plausible bug" — no bug in this app can make it fail. Inline the `await` at line 662; the comment at 648-650 already carries the reasoning. Drop the test.

### 6. The probe sends `Accept: application/json`

Inherited from `PlexConfig.headers` (`plex_config.dart:72`); mpv sends no such thing. If Plex ever content-negotiates, `data` decodes to a Map, `.toString()` never starts with `#EXTM3U`, and the loop polls the whole window. Pass `Accept: */*` alongside the existing `Range` header so the probe mirrors the player's request shape.

## Nits

- `plex_client.dart:2748` logs `${timeout.inSeconds}s` → "did not become ready within 0s" for sub-second timeouts.
- The test comment at `plex_playback_data_request_test.dart:796-800` cites 12,927 requests in a 2 s window; the PR body says 3,651 in 600 ms. Both describe a busy-loop that never existed on `main`.
- That same test burns 600 ms of wall clock and asserts `inInclusiveRange(3, 8)`. AGENTS.md asks for `fakeAsync`/injected clocks over wall-clock timing; the bound is defensible but fragile under a loaded runner.
- The segment probe goes through `_send`, which UTF-8-decodes the body into a `String`. With `Range: bytes=0-0` that's one byte; if a PMS ignores `Range`, it's a full segment through `tryIsolateRun`. `getBytes` sidesteps it.
- `transcodeOffset` is a backend-neutral option only Plex reads; Jellyfin's `StartTimeTicks` equivalent stays unwired. Fine as scope — worth saying so in the doc comment at `playback_initialization_types.dart:64-67`.

## What checks out

- `selectReadinessProbeTarget` boundary math is right: `cumulative > offset` returns the segment whose window *contains* the offset, including the exact-boundary case (2.0 s into 2 s segments → `00001.ts`).
- The offset-equals-open-position invariant holds at all three construction sites. Initial open: `_resolveOpenResumePosition` with no `requested` returns `metadata.viewOffsetMs`, the same value passed at `video_player_screen.dart:1276`. "Play from Beginning" goes through `copyWith(viewOffsetMs: 0)` (`media_context_menu.dart:702`), and `offset > Duration.zero` drops the param, so it stays offset-free. Reload passes the resolved resume position directly.
- Moving `_resolveOpenResumePosition` ahead of the resolve preserves the old `_offlineLibraryMode || result.isOffline` semantics via the re-resolve at `episode_navigation.dart:690-700`; costs one extra local read only when a download wins.
- `stream-lavf-o` is untouched (`playback_open.dart:569-573`), as claimed.
- `_http.get` genuinely does not throw on non-2xx, so the "non-2xx is the not-ready signal" premise is sound.
- Live TV is unaffected: it has its own tune path and the in-place reload is `routeKind: vod`.

One assumption stays unguarded rather than wrong: "the offset playlist still covers the full title from segment zero" is verified on one PMS. If a build ever returns an offset-relative playlist, `selectReadinessProbeTarget` picks a segment past the end and every open eats the full window. With the throw removed (blocking #3) that degrades to today's behavior instead of a hard failure, which is the main reason I'd rather the probe never be fatal.

## Verification run

`flutter test` on both touched files (51 pass) and all of `test/screens/video_player` (97 pass); `dart run scripts/check_analyzer.dart` clean; `dart format --set-exit-if-changed` clean on all 7 files. Blocking findings 1 and 2 come from two scratch tests against `PlexClient.waitForTranscodeReady`, since deleted. I did not reproduce the live-PMS playlist measurement — that assumption rests on the author's single-server observation.

toluLikesToCode:
All six plus the nits, in 77d6da8e.

**The probe no longer fails an open.** Both throw sites are gone; the call sites await and proceed unconditionally, so mpv sees what the server is actually doing and the log-stream classification runs as it does today. Inside the loop a 500 returns immediately without classifying, and cancellation breaks instead of sleeping. Each of your three scratch measurements has a regression test now, and I verified all three red-first by reverting the fixes: the failover test counts 4 all-endpoints-exhausted signals under the old code inside a 200ms window, the cancellation test only completes at the window end, the 500 test polls instead of handing off.

**Failover needed two changes, not one.** The playlist leg passes `allowEndpointFailover: false`. The segment leg couldn't take `getBytes`: its consume only logs the status (`media_server_http_client.dart:172-176`), so the probe would be blind to 404-vs-500-vs-2xx. Added `getRaw` — bytes plus status, never text-decoded, non-2xx returned rather than thrown — which also settles the decode nit, since a PMS that ignores `Range` no longer routes a full segment through `tryIsolateRun`. It lives on the base client, so it never enters the failover path at all; that is why the regression test 503s the playlist rather than the segment. Asserting zero exhausted signals meant plumbing `onAllEndpointsExhausted` through `PlexClient.forTesting`, mirroring the Jellyfin fixture.

**404 keeps polling at every depth.** I have no trace saying manifest-404 is terminal, so the loop treats it as not-ready for the whole window and mpv classifies whatever survives.

**The offset gate is folded into `waitForTranscodeReady`** — a URL without an offset returns true with zero requests, tested, so the public method can no longer probe segment zero. What remains at the two call sites is a guarded await whose staleness tokens differ (`attempt.isCurrent` vs `isCurrentReload()`); that is why I stopped short of a shared helper. Say so if you want one extracted anyway.

**`resolveReplacementAfterStoppedPlayback` is inlined** and its test dropped; the comment above the stop future carries the reasoning.

**Probe requests send `Accept: */*`** on both legs, asserted alongside the `Range` assertions, and the timeout log reports milliseconds.

On the remaining nits: the bounded test runs under `fakeAsync` with an exact cadence — the playlist hop, then segment polls at 0/50/100/150/250/450ms, 7 requests rather than a range — which meant moving the probe's deadline onto `clock.now()`; `clock` graduates from transitive to a direct dependency at its locked 1.1.2. You're right that neither busy-loop figure ever shipped on main; the comment now attributes 3,651 and 12 to the defective drafts they were measured against. `transcodeOffset`'s doc now says Jellyfin's StartTimeTicks equivalent is intentionally unwired.

Full suite green (5,550 tests), analyzer and format clean.

edde746:
Reviewed `77d6da8e` against base `f63d0fe4`, in a clean worktree.

All six items and the nits are addressed, and I re-derived the three regression tests' cadences rather than trusting their assertions — they encode the right contracts. But the immediate-500 handoff has a hole on the exception path, and it comes from a premise I signed off on last round, so that one is on me as much as on the patch.

*(Edited after posting: finding 1 below, and the corresponding correction to "What checks out". The original version of this comment repeated the false premise.)*

## Blocking

### 1. A 500 that fails JSON decode is polled for the full window, defeating the fast-path this commit added

The whole of last round's #3 rested on a claim I made and this PR's doc comment now repeats at `plex_client.dart:2689`: *"Other non-2xx responses are the expected not-ready signal — `_http.get` does not throw on them"*. That is false. `_send`'s consume decodes the body **before** building the response, and a decode failure throws a **status-bearing** `MediaServerHttpException` (`media_server_http_client.dart:302-313`). `MediaServerHttpException.from` returns an existing instance unchanged (`media_server_exceptions.dart:59`), so it arrives at the probe's `on MediaServerHttpException` catch (`plex_client.dart:2769`) with `statusCode == 500` — where it is counted as a transport failure and polled.

`_decodeBody` only throws when the content-type contains `json` and `jsonDecode` fails on a non-empty body (`media_server_http_client.dart:458-473`), so the trigger is a 500 that declares JSON and returns something else — a proxy, relay, or gateway error page in front of PMS is the realistic source, not PMS itself. Narrow, but the shipped behavior contradicts the method's own documented contract, and the fix is four lines.

Two scratch tests against `waitForTranscodeReady`, `timeout=600ms, poll=50ms`, both legs 500:

```
content-type: text/html            → 1 request,  "handing off on HTTP 500"     ← contract
content-type: application/json     → 6 requests, full 600ms window             ← observed
  [D] Plex transcode readiness probe transport failure  ERROR: MediaServerHttpException(unknown: HTTP 500: Failed to decode response body: FormatException…)
  [W] Plex transcode did not become ready within 600ms (playlistDepth=0, lastStatus=none, consecutiveFailures=6)
```

At production defaults that is 15 s in front of the server-limit dialog — the precise delay `playback_open.dart:563-565` exists to prevent, and the precise regression last round blocked on, reintroduced through the door nobody checked. Note `lastStatus=none` in that warning: the throw path never reaches `lastStatus = statusCode` (`:2741`), so the one diagnostic that would have shown a 500 reports the server as silent.

Mirror the response branch inside the catch, which also repairs the log:

```dart
} on MediaServerHttpException catch (e) {
  if (e.isCancellation) return false;
  lastStatus = e.statusCode ?? lastStatus;
  if (e.statusCode == 500) {
    appLogger.i('Plex transcode readiness probe handing off on HTTP 500');
    return false;
  }
  consecutiveFailures++;
  …
```

Better still, hoist the decision so the two paths cannot drift again — they already have once. And correct `:2689`: `_http.get` does not throw *on the status*, but it does throw *carrying* one.

Regression test: playlist leg answers 500 with `content-type: application/json` and an unparseable body; assert one request and no poll wait, alongside the existing `text/html` case. That pins both paths to the same rule.

## Should fix

### 2. The playlist-shorter-than-offset fallback probes the last segment — which, by this PR's own premise, seeks the session

`plex_client.dart:2646` documents it and `:2671` implements it: `return (sawSegmentDurations ? lastEntry : firstEntry) ?? ''`. Put that next to the two statements the design rests on — "probing the first entry would steer the transcoder back to the start — requesting a segment is how a client seeks a Plex HLS session" (`:2643-2644`, repeated at `:2686`) — and the fallback does the exact thing the mechanism exists to avoid, in the one case where it has no idea where the offset lives.

It's also worse than last round's read of it ("picks a segment past the end and every open eats the full window"). It doesn't eat the window: it *succeeds*. Scratch test feeding the offset-relative playlist shape the PR explicitly flags as unverified (resume 8,062 s, playlist `08062.ts`–`08121.ts`):

```
GET …/video/:/transcode/universal/start.m3u8?offset=8062.000000 → 200
GET …/video/:/transcode/universal/08121.ts                      → 206
ready = true
```

The probe reports the session is positioned at the resume point, having just asked it to produce ~60 s past it; mpv then opens and requests the first segment, relocating it a second time. A resume that stutters instead of one that is merely slow.

Under the full-title playlist you measured this is unreachable — `openResumePosition` is bounded by the title duration at all three construction sites — so this is an unguarded assumption, not a live bug. But it's the assumption you already called out as resting on one PMS, and the degradation you argued for last round ("that degrades to today's behavior instead of a hard failure") is not what the code does.

Make it a hand-off rather than a guess: when `sawSegmentDurations` is true and the walk finished without crossing `offset`, the playlist cannot tell you where the offset is, so return `''` and let the loop fall through to `// A manifest with no media entries yet` (`:2756`) — no segment is touched, the window expires, mpv opens into today's behavior. `lastEntry` then has no readers and comes out; `firstEntry` stays for the master-playlist descent. If you'd rather not spend the window polling a manifest that will never answer, returning `true` from `waitForTranscodeReady` on that condition is equally defensible — the probe is best-effort, and a probe aimed at the wrong segment is worse than no probe.

### 3. The probe serializes ahead of every other pre-open step instead of hiding behind them

`playback_start.dart:156-162` runs before `_commitPlaybackSession` (`:180`), `SettingsService.getInstance()` (`:185`), `_prepareFrameRateForOpen` (`:202`) and `_primeDisplayCriteria` (`:245`); the open is at `:268`. `episode_navigation.dart:687-694` sits ~90 lines ahead of `_openMediaOnPlayer` (`:778`), before the subtitle resolve, the settings load, the frame-rate plan (which on Android holds for an actual display mode switch) and `attempt.trackMutationDrain`.

Nothing between the probe and the open needs the session to be ready, so every millisecond the probe spends waiting is added to time-to-first-frame rather than overlapped with work that has to happen anyway. Moving the `await` and its staleness check to immediately before `_openMediaOnPlayer` in both flows costs nothing and makes the guarantee fresher at the only moment it means anything — when mpv attaches. On the reload path it also shrinks the window in which a not-yet-open replacement session sits idle.

## Nits

- `getRaw`'s doc (`media_server_http_client.dart:180-183`) contrasts it with `getBytes` and `get` but omits the property the probe actually depends on: it exists only on the base client, so unlike `get` it can never enter the failover cascade (`failover_http_client.dart:89-135` overrides `get` alone). That's why the segment leg needs no `allowEndpointFailover: false`, and right now it's recorded in a PR comment rather than in the code. It is also why the segment leg is immune to finding 1 — `getRaw` never decodes — which is worth stating rather than leaving to be rediscovered.
- `getRaw` diverges from `_send` by not deriving `effectiveUri` from `http.BaseResponseWithUrl` (`:199-204` vs `:294-297`, `:320`). Harmless here — the probe only reads `statusCode` — but a trap for the next caller that follows a redirect.
- `getRaw` buffers the whole body into a `data` its only caller discards. One byte with `Range: bytes=0-0`; a full segment in memory from a PMS that ignores `Range`. `streamed.stream.drain()` is the honest shape for a probe, or keep the bytes and say what they're for.
- `transcodeStreamOffsetFromUrl` (`:2633`) matches on `uri.path`, which preserves percent-encoding, so the `%3A` spelling of the same URL returns null and silently switches the probe off. Not reachable today — `transcodeUrl` at `:3462` is string concatenation with a literal colon — but a silent-off failure mode is a bad one to leave to luck.
- The probe takes no `AbortController`, so a superseded open (`!attempt.isCurrent` / `!isCurrentReload()`) leaves it polling the PMS for the balance of its 15 s window. Bounded and harmless, but both call sites already hold a staleness token and could pass an abort.
- A resumed transcode that 404s for real now reaches `mediaUnreadableDialog` 15 s later than on `main`. That is the trade we agreed to, but the doc comment at `:2674-2695` records only the 500 fast-path rationale. One sentence on the 404 delay would stop the next reader from "fixing" it.
- Both call sites discard the `bool`. That is the design; worth noting it now exists only for the tests.
- No test covers the screen-level wiring — that `transcodeOffset` equals the position the player opens at. It holds structurally today (below), and the seam is expensive to reach, so I'm not asking for one; just flagging that the PR's central invariant is verified by reading, not by CI.

## What checks out

- All three new regression tests encode real cadences. Re-derived by hand: the bounded test's 7 requests are exactly the playlist hop plus segment polls at 0/50/100/150/250/450 ms under `timeout=600, poll=50` with the `cf > 3 → 1 << clamp(cf-3, 0, 2)` backoff; the cancellation test's `requestCount == 2` proves the third request throws before sending, costing one poll interval rather than the window; the 500 test completes under `flushMicrotasks` with no timer at all — on the response path, which is the half finding 1 says is only half the rule.
- Failover bypass is real on both legs, not just asserted: `FailoverHttpClient` overrides `get` only (`failover_http_client.dart:89-135`), so `getRaw` structurally cannot cascade, and the playlist leg passes `allowEndpointFailover: false` (`plex_client.dart:2728`).
- **Correction to my last review.** I wrote that "`_http.get` genuinely does not throw on non-2xx, so the 'non-2xx is the not-ready signal' premise is sound", and this comment originally repeated it for `_perform`. `_perform` itself does not throw on status (`media_server_http_client.dart:331-372`) — that part holds, and it is what makes `getRaw`'s returned status sound — but `_send`'s **consume** does, carrying the status, on a JSON-declared body that will not parse. Finding 1 is the consequence.
- The offset-equals-open-position invariant survives the reorder, by construction rather than by coincidence. `_playbackOpenTiming` passes `resumePosition` through untouched (`video_player_screen.dart:318-327`) into `Media(videoUrl, start: timing.mediaStart)` (`playback_open.dart:639`) — nothing clamps a near-end or near-zero resume, so `offset=` and mpv's `start:` are the same number. Initial open: `_currentMetadata` is assigned once at `video_player_screen.dart:876` and mutated only by reload, so `:1276` and `playback_start.dart:238` read the same `viewOffsetMs`; the `_isOfflinePlayback` branch inside `_resolveOpenResumePosition` cannot diverge them because the transcode branch returns `isOffline: false` (`plex_client.dart:3469`).
- The stop-ordering fix survives supersession, which is the case that worried me about pulling the await forward. `episode_navigation.dart:646` can return between `_sendStoppedProgressOnce()` and the `await`, abandoning the future — but `PlaybackProgressTracker.sendStoppedProgressOnce` memoizes (`playback_progress_tracker.dart:247-249`), so the next reload awaits that same in-flight stop instead of racing it, and `_sendStoppedProgressOnce` already `.catchError`s (`video_player_screen.dart:2166`), so the abandoned future cannot surface as an unhandled error.
- `resolveReplacementAfterStoppedPlayback` and its test are gone: `player_initialization_lifecycle_test.dart` is net-unchanged against base.
- The offset gate really is inside the method — no-offset URLs return with `requestCount == 0`, so the public method can no longer probe segment zero. `Duration.zero` and "Play from Beginning" (`media_context_menu.dart:702`) stay offset-free through `offset > Duration.zero` (`:2919`).
- Backoff math matches the doc: first extra delay at the 4th consecutive failure, capped at 4x (2 s at the production interval).

## Verification run

Clean worktree at `77d6da8e`:

- `flutter test` — 5,550 pass, 5 skipped.
- `flutter test test/services/plex_playback_data_request_test.dart test/screens/video_player` — 132 pass.
- `dart run scripts/check_analyzer.dart` — clean (needed `flutter pub get` in `packages/wakelock_plus` and `packages/saf_util` first; worktree setup, not the PR).
- `dart format --set-exit-if-changed --line-length 120` on all 8 touched Dart files — clean.

Findings 1 and 2 come from scratch tests against `waitForTranscodeReady`, since deleted. I did not reproduce the live-PMS playlist measurement; that assumption still rests on your single-server observation, which is exactly why finding 2 is worth the four-line guard.

toluLikesToCode:
All three plus the nits, in 2961b178.

The 500 decision is one closure now, applied to responses and status-bearing exceptions alike, and the catch records `lastStatus` so the timeout log can't call a 500 silent anymore. Your JSON-decode test case is in, and I checked it red-first — with the catch-path hand-off removed it's still polling when microtasks drain.

For the short-playlist fallback I took the skip: `selectReadinessProbeTarget` returns null when the durations never cross the offset, and the probe returns true without touching a segment. Your own line settled it — a probe aimed at the wrong segment is worse than no probe. Test feeds the 08062–08121 shape from your scratch run and asserts no `.ts` request goes out.

The probe moved to just before `_openMediaOnPlayer` in both flows, through one shared helper. That helper also picks up the abort nit: a new probe aborts the previous one, dispose aborts the last, and a pre-aborted controller never issues a request.

Rest of the nits: `getRaw` became `getStatus` — drains the body, derives `effectiveUri` like `_send`, doc now says why it's failover-proof and decode-proof. The URL matcher compares decoded segments so `%3A` parses. The 404 delay and the tests-only bool are documented.

One admission: fixing the fallback broke three of round two's fixtures — a one-second playlist with `offset=495` meant the probe now skips itself before reaching the behavior under test. They'd been leaning on the exact guess you flagged. Offsets corrected; cadences unchanged.

5,553 green, analyzer and format clean.

edde746:
Landed on main as 8740a19f3 (squash of 2961b178, conflicts with 7437b4320's forTesting signature resolved by keeping main's superset). Verified post-merge: full suite 5,716 pass, analyzer parity clean, and a live-PMS smoke of the #1840 scenario — offset=2848.512000 decision/start, readiness probe answered 206 on the first poll, and real mpv with --start=2848.512 opened 02848.ts with zero 404s, playback restart complete @ 2848.95s.

PR #1828, opened 2026-08-07
fix(auth): fall back to xdg-open when gio launch fails on Linux

### LINKED ISSUE #1477: Linux: Cannot "Sing in with Plex"
### Bug Description

When choosing the "Sing in with Plex" option, I get this error: `Plex PIN auth failed  ERROR: PlatformException(Launch Error, Failed to launch URL: Failed to execute child process “gio-launch-desktop” (No such file or directory), null, null)`

### Steps to Reproduce

Click the "Sing in with Plex" button on Fedora KDE

### Expected Behavior

To start the sing in process

### Actual Behavior

Throws an error

### Plezy Version

2.8.0, 1.35.2

### Platform

Linux

### Media Server

Plex

### Device Information

Fedora 44, KDE Plasma 6.7.1

### Logs and Stack Traces

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

vavallee:
## What

Fixes #1477. On some Linux desktops, "Sign in with Plex" fails with:

```
Plex PIN auth failed  ERROR: PlatformException(Launch Error, Failed to launch URL:
Failed to execute child process "gio-launch-desktop" (No such file or directory), null, null)
```

## Why

`url_launcher` opens URLs on Linux through GLib, which execs the `gio-launch-desktop` helper. That helper is not present on every desktop (reported on Fedora/KDE without the GTK stack), so `launchUrl` throws. In the Plex PIN flow both the primary (`inAppBrowserView`) and the existing external-browser fallback route through the same gio path, so the error propagates to the flow's catch and shows an unrecoverable "Plex PIN auth failed" message — the user can never sign in.

## Change

- Add `lib/utils/external_url_launcher.dart` with `launchExternalUrl`, which retries via `xdg-open` (desktop-agnostic, shipped by xdg-utils) when the primary launch fails on Linux. On non-Linux a thrown launch is rethrown, preserving `launchUrl`'s existing contract.
- Route the sign-in's external-browser fallback in `plex_pin_auth_flow.dart` through it. Behavior on macOS/Windows/mobile/TV is unchanged.

The helper is written with injection seams so its branching is unit-tested without a real launcher.

## Tests

`test/utils/external_url_launcher_test.dart` covers: primary success skips the fallback; the requested `LaunchMode` is forwarded; a Linux throw (gio missing) and a Linux `false` return both fall back to xdg-open; non-Linux rethrows/returns false without invoking the fallback.

`flutter analyze`, `dart format`, the `dart_code_linter` unused-code/unused-files checks, and the existing `plex_pin_auth_flow` / `auth_screen` tests pass locally. I don't have a Fedora/KDE machine to reproduce the original gio failure end-to-end; the fallback path is covered by the injected-launcher tests.

### DIFF
--- a/lib/screens/auth/plex_pin_auth_flow.dart
+++ b/lib/screens/auth/plex_pin_auth_flow.dart
@@ -10,6 +10,7 @@ import '../../services/plex_auth_service.dart';
 import '../../focus/focusable_button.dart';
 import '../../theme/mono_tokens.dart';
 import '../../utils/app_logger.dart';
+import '../../utils/external_url_launcher.dart';
 import '../../utils/platform_detector.dart';
 
 /// Whether the Plex PIN hand-off must stay in-app as a QR code instead of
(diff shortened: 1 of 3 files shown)


edde746:
Wrong approach, #1477 has been fixed.

PR #1834, opened 2026-08-08
Add SteamOS environment variables for plezy

ItsRetroPup:
Added SteamOS fixes for scaling and client-side decorations.


aldobarr:
This would force all Linux users on 2x scaling which would look oversized on desktop environments. You should instead change your launcher to add the 2 env variables. I.e. in steam's launch options:
`env GDK_SCALE=2 GTK_CSD=0 %command%` and that's it.

or if you use a .desktop launcher, change Exec line to:
`Exec=env GDK_SCALE=2 GTK_CSD=0 plezy`

No change to builds necessary.

PR #1836, opened 2026-08-08
fix(plex): carry embedded subtitles in the transcode

toluLikesToCode:
During a transcode with direct play disabled, an embedded subtitle has no fetchable URL — `/library/streams/{id}.{ext}` answers 501 for embedded streams, and the raw part file answers 503, the very thing the transcode exists to avoid. The client still tried to read the source container as a sidecar, ffmpeg's auto-reconnect retried the 503 forever, and the open guard tore down every subtitle with "Selected subtitles could not be loaded". This is the Plex half of #1738.

The fix asks the server to deliver the selected embedded stream: text tracks ride a segmented WebVTT rendition, image tracks are burned in (HLS has no bitmap rendition), keyed sidecar files stay client-fetched. ffmpeg's HLS demuxer never applies `X-TIMESTAMP-MAP` (`webvttdec.c` passes cue times straight through; `hls.c` has no option for it), so the declared offset is read from the transcode's own segments at open and compensated via `sub-delay` — the viewer's sync slider rides on top and still reads zero. `sub-clear-on-seek` on transcode opens stops re-delivered cues stacking after a seek. A second commit stops a quality-capped transcode re-encoding profile-compatible audio (EAC3 5.1 640k was becoming 360k AAC): the video cap rides the client-profile `add-limitation` and `directStreamAudio=1` copies the audio through.

Everything here was measured against a live PMS 1.43 rather than inferred — including the surprise that cue-less spans are served as bare `WEBVTT` stubs with no timestamp map, which the offset probe has to skip. Verified on-device across three rounds: no stall or error, subtitles in sync including after seeks, audio copied.

I'll be upfront that this isn't the ideal shape: it leans on the transcoder where direct play/direct stream just hand the player the subtitle stream to handle natively. That path isn't reachable here — the source container is the one URL a disabled direct play forbids — and this is the best mechanism I found that actually works. If anyone sees a better one, I'm happy for it to win. The Jellyfin half of #1738 is deliberately not in this PR; it's a separate defect (the sidecar loads but the track is never applied) and will follow as its own PR so the two backends end up consistent. Please review carefully.

### DIFF
--- a/lib/screens/video_player/parts/build.dart
+++ b/lib/screens/video_player/parts/build.dart
@@ -293,6 +293,7 @@ extension _VideoPlayerBuildMethods on VideoPlayerScreenState {
                         selectedQualityPreset: _selectedQualityPreset,
                         serverSupportsTranscoding: _serverSupportsTranscoding,
                         isTranscoding: _isTranscoding,
+                        subtitleTimelineOffsetMs: _subtitleTimelineOffsetMs,
                         isOfflinePlayback: _isOfflinePlayback,
                         sourceAudioTracks: sourceAudioTracks,
                         selectedAudioStreamId: _selectedAudioStreamId,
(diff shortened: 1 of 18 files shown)


edde746:
Closing in favor of https://github.com/edde746/plezy/commit/f4ce60611b41031e4e4025b85fab4a31fa4a4ab6

PR #1841, opened 2026-08-09
feat(rpm): add rpm signing

aldobarr:
@edde746 I'm creating this as a draft because my motivation for this is to see if you'd be willing to setup an RPM repository which would allow RPM package managers to auto update Plezy.

This PR sets up some of what you would need for that, signing the RPMs. The other part would be hosting the repository metadata and setting up the signing keys etc. The metadata is relatively small, it would mostly just be the list of versions and links to the signed RPM assets, fingerprints, public key etc.


edde746:
I’m going to decline this. Linux packaging is too fragmented, Plezy has very few Linux users, and supporting an official repository for one package format would create an expectation to support the others as well. I don’t use Linux or have an interest in maintaining that packaging surface, so the maintenance cost isn’t justified.

aldobarr:
> I’m going to decline this. Linux packaging is too fragmented, Plezy has very few Linux users, and supporting an official repository for one package format would create an expectation to support the others as well. I don’t use Linux or have an interest in maintaining that packaging surface, so the maintenance cost isn’t justified.

Fair enough, I'll submit a PR to your readme to add a link for rpm users to my unofficial repository I'll set up instead if you want that.

PR #1845, opened 2026-08-09
fix(plex): keep compatible audio on capped transcodes

toluLikesToCode:
`maxVideoBitrate` budgets the whole stream rather than the video track, so any
capped quality preset dragged profile-compatible audio down with it. Measured
against Plex 1.43, a 720p/3 Mbps session on a source carrying EAC3 5.1 at 640
kbps came back as 360 kbps AAC — the audio was re-encoded only to fit a budget
the video was already being fitted to.

The cap does not need that parameter. `_buildPlexHlsClientProfileExtra` already
sends the preset's ceiling as a client-profile limitation
(`add-limitation(scope=videoCodec&scopeName=*&type=upperBound&name=video.bitrate&value=<kbps>&replace=true)`),
which constrains the video track on its own. With the redundant
`maxVideoBitrate` dropped, `directStreamAudio=1` lets the server copy the source
audio whenever its codec is one the profile declares in the HLS transcode target
(`aac`, `ac3`, `eac3`, `mp3`). Audio the target cannot carry — TrueHD, DTS,
FLAC, Opus — still transcodes exactly as before, and the per-preset
`directPlay`/`directStream` pinning is untouched.

Worth stating plainly: the preset now bounds the video track alone, so total
stream bandwidth can exceed the nominal figure by the source audio bitrate. That
matches what the picker already implies — the field is `videoBitrateKbps`, and
`qualityPresetSizeEstimate` adds an audio allowance on top of it rather than
carving one out of it — but two things now trail the behaviour, and I have left
both alone rather than widen this change:

- the size estimate assumes a flat 192 kbps of audio, which under-reports a
  direct-streamed 5.1 track;
- the Jellyfin path still sends the preset as `maxStreamingBitrate`, a
  whole-stream budget, so the same row means slightly different things per
  backend.

Happy to fold either in here if you would rather they moved together.

Covered by a test asserting that a capped preset emits no `maxVideoBitrate`,
carries `video.bitrate` in the client profile, and sets `directStreamAudio=1`.


edde746:
This PR should be ~2 lines.

toluLikesToCode:
> This PR should be ~2 lines.

Sure thing, I'll just remove most of the comments and the test

PR #1848, opened 2026-08-09
Add unofficial rpm package manager instructions

aldobarr:
Adding my unofficial rpm repository to the package manager sections for if you want it.


edde746:
Implemented in https://github.com/edde746/plezy/commit/e59b79266afe6941cc69cb5c7b686be5bd756f13

PR #1894, opened 2026-08-12
Add a play from beginning button to the detail screen

slevin-7:
On the show detail screen the play button resumes the on-deck episode, but there is no way to restart that episode from the start. "Play from beginning" only exists in the ⋮ menu for movies - the show item itself never carries a view offset, so the menu entry never appears for shows.

My use case: I fall asleep to a show most nights, so the next evening I want to rewatch the current episode from the start instead of resuming at wherever I dozed off.

This adds a replay button next to play, shown only when there is something to restart: the on-deck episode with progress for shows, or the item's own view offset for movies and episodes. It plays through the existing play-from-start path (view offset cleared plus `resolveWatchState: false`, same as the context menu entry does). `navigateToVideoPlayerWithRefresh` now passes `resolveWatchState` through to make that possible.

Reuses the existing `mediaMenu.playFromBeginning` string, so no new translation keys.

On narrow phone layouts the button drops out of the compact tiers like the other secondary actions - the episode card context menu still covers restarting there.

### DIFF
--- a/lib/screens/media_detail/action_buttons.dart
+++ b/lib/screens/media_detail/action_buttons.dart
@@ -53,6 +53,32 @@ extension _MediaDetailActionButtons on _MediaDetailScreenState {
       }
     }
 
+    // Restart target: the item whose in-progress playback a "play from
+    // beginning" would discard. For shows that is the on-deck episode (the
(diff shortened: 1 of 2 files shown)


edde746:
Clutter that duplicates functionality.

Issue #1796, opened 2026-08-05
Media without logos/banners clip text with ellipsis too early on wide displays


JackCuthbert:
### Bug Description

When media doesn't have a logo/banner image, the title text clips very early despite being on a wide display that could fit the whole thing.

I'm assuming this behaviour is because logos/banners are expected on _most_ media so this clipping doesn't usually happen. Filed as a bug but could definitely be classified as a feature request to support media without logos.

### Steps to Reproduce

View any media info page that lacks a logo/banner image and has a long title.

### Expected Behavior

Text runs the full length of the available display, wraps to new lines, or decreases in size to fit.

### Actual Behavior

Text is clipped with ellipsis very early on the second line.

### Plezy Version

2.11.0

### Platform

iOS, macOS

### Media Server

Plex

### Device Information

iPad Pro 12.9", MacBook Air 13" M4

### Screenshots

Problematic video where the second line of text is clipped despite having real estate for the whole title:

<img width="1362" height="872" alt="Image" src="https://github.com/user-attachments/assets/b27fa383-09e8-44b7-9c50-7540f8c33763" />

On media that lacks a banner but text _does_ fit within the constrained width it looks OK:

<img width="1362" height="872" alt="Image" src="https://github.com/user-attachments/assets/bcbd93f5-7e59-486d-8a41-7674c28cb9f9" />

PR #1913, opened 2026-08-13
fix(ui): let logo-less detail titles use full hero width

vavallee:
## What

When a movie or show has no clear-logo image, the detail hero falls back to rendering its title as text. That fallback was boxed at the *logo's* width cap — 400px on the phone/desktop hero, 790px on the TV hero — so `FittingTitleText` shrank the font and ellipsized a long title inside the logo box even on a wide display that had plenty of room. This is #1796 (reported on iPad Pro 12.9" / MacBook Air, Plex). At 40px title font, the 400px box clips titles as short as ~33 characters.

## Fix

One rule, applied to both heroes: **a logo-less title may use up to twice the logo box width, never more than the hero.** `_buildDetailLogoOrTitle` takes an optional `titleWidth`; the text fallback uses it, while a clear-logo *image* keeps the narrower `width` cap. `titleWidth` defaults to `width`, so logo-present paths are unchanged.

Per layout:

| Layout | Effective title width |
|---|---|
| Phone | the hero (narrower than the cap) — unchanged |
| Tablet / desktop | 800px (2 × the 400 logo box) — wider than before, but a title column, not edge-to-edge |
| TV | the foreground column (the right 43% of the screen is already reserved for artwork, which bounds it below the scaled 2× cap) |

With the 800 cap, titles up to ~70 characters fit on two lines at full size (measured with the bundled font at 40px — "Dr. Strangelove or: How I Learned to Stop Worrying and Love the Bomb" fits). Longer outliers still ellipsize at the cap, because `FittingTitleText` only shrinks the font on height pressure, not width; if you'd like those to shrink-to-fit instead, I can follow up with an opt-in width-aware shrink.

## Testing

- `wide non-TV hero gives a logo-less title a capped width wider than the logo box`: 1400×900 non-TV viewport, asserts the title box is exactly 800 — wider than the old 400, not the 1368px hero. Fails with the cap removed (box becomes 1368).
- `phone-width hero keeps a logo-less title inside the hero`: 390×844, asserts the box is exactly the hero width (358) — the "mobile unchanged" claim; fails if the cap were ever narrower than the hero.
- `TV hero gives a logo-less title more than the logo box, within its column`: 1920×1080 TV, asserts the box exceeds the 790px logo box and stays inside the foreground column. Fails on the pre-fix behavior (box = 790).
- Existing `TV detail scales fallback title to fit logo bounds` still passes.
- `flutter analyze`, `dart format --line-length 120`, `dart_code_linter` unused-code/unused-files, full `flutter test` (5825): clean.

Screenshots from the Windows desktop build are in the comments.

### DIFF
--- a/lib/screens/media_detail_screen.dart
+++ b/lib/screens/media_detail_screen.dart
@@ -3441,6 +3441,7 @@ class _MediaDetailScreenState extends State<MediaDetailScreen>
                                 context,
                                 metadata,
                                 width: logoWidth,
+                                titleWidth: constraints.maxWidth,
                                 height: logoHeight,
                                 titleBuilder: (context, title) => _buildDetailTitle(
                                   context,
@@ -3585,11 +3586,16 @@ class _MediaDetailScreenState extends State<MediaDetailScreen>
     required double width,
     required double height,
     required Widget Function(BuildContext context, String title) titleBuilder,
+    double? titleWidth,
   }) {
     Widget titleFallback(BuildContext context) => titleBuilder(context, metadata.displayTitle);
 
     if (metadata.clearLogoPath == null) {
-      return SizedBox(width: width, height: height, child: titleFallback(context));
+      // A clear-logo image is capped at [width] so it doesn't stretch across a
+      // wide hero, but the text-title fallback for logo-less media wants the
+      // full available width — otherwise FittingTitleText shrinks and clips a
+      // long title inside the logo box even when the display could fit it (#1796).
+      return SizedBox(width: titleWidth ?? width, height: height, child: titleFallback(context));
     }
 
     return SizedBox(
@@ -4209,6 +4215,7 @@ class _MediaDetailScreenState extends State<MediaDetailScreen>
                           context,
                           metadata,
                           width: logoWidth,
+                          titleWidth: constraints.maxWidth,
                           height: logoHeight,
                           titleBuilder: (context, title) => _buildDetailTitle(
                             context,
--- a/lib/screens/media_detail_screen.dart
+++ b/lib/screens/media_detail_screen.dart
@@ -3412,6 +3412,12 @@ class _MediaDetailScreenState extends State<MediaDetailScreen>
             actionGap +
             actionHeight;
         final logoWidth = desiredLogoWidth < constraints.maxWidth ? desiredLogoWidth : constraints.maxWidth;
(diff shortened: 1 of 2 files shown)


edde746:
When making UI changes, please provide screenshots.

vavallee:
Screenshots from the Windows desktop build (wide window). To make the case reproduce deterministically I forced a logo-less item with a long title, so it's the same item and the same window in both shots — only the title layout differs. The hero backdrop is empty because the sample item has no artwork.

**Before** — the title is boxed at the ~400px logo-width cap and clips on line 2:

before

**After** — the title uses the full hero width:

after

edde746:
I think making it wider is probably right, however, I don't think making it full width is great. Would also need consideration for how to handle it on TV and mobile.

vavallee:
Agreed — full width reads as a banner. Pushed a revision (26862879) that bounds it instead: **the title may use up to twice the logo box, never more than the hero.**

What that means per layout:

- **Mobile** — the hero is narrower than the cap, so the title just fills the hero, exactly as it does today. No visible change.
- **Tablet / desktop** — 800px (2 × the 400px logo box). Wider than the old cap, but a title column rather than edge-to-edge. Measured with the bundled font at the 40px title size, that fits titles up to ~70 characters on two lines — "Dr. Strangelove or: How I Learned to Stop Worrying and Love the Bomb" fits at full size — where the old 400px box clipped titles as short as ~33 characters.
- **TV** — same rule with the scaled logo box, though in practice the TV foreground is already a bounded column (the right 43% of the screen is reserved for artwork), and that column stays under the cap, so the title uses the column.

Each of these is pinned by a widget test (exact 800 box on a wide desktop viewport, exact hero width on a phone viewport, wider-than-logo-box within the column on a 1080p TV viewport).

One edge to be transparent about: titles longer than ~70 characters (e.g. Borat's full title) still ellipsize at the cap, because `FittingTitleText` only shrinks the font under height pressure, not width. If you'd rather those shrink a few px to fit, I can add an opt-in width-aware shrink to `FittingTitleText` as a follow-up.

Screenshots (Windows desktop build, same forced logo-less item):

**Desktop, maximized (~2000px window)** — title column at the 800 cap:

desktop wide

**Desktop, narrower window (~1170px)** — still bounded, adapts:

desktop narrow

**TV layout** — title inside the foreground column:

tv

PR #1925, opened 2026-08-14
fix(ios): pin the AVFoundation audio output so multichannel survives

toluLikesToCode:
Every multichannel source plays in stereo on iPhone and iPad, and the cause is
which audio output driver mpv picks. `audio_out_drivers[]` lists `audiounit`
ahead of `avfoundation` and the iOS slice ships both, so `audiounit` wins every
PCM probe. Its init clamps outright — `ao->channels = MP_CHMAP_INIT_STEREO`
when `instance.outputNumberOfChannels <= 2` — and that branch is always taken,
because nothing on this path calls
`AVAudioSession.setSupportsMultichannelContent:`, whose documented default is
`NO`. The session never offers more than two channels, so mpv inserts a
swresample downmix ahead of the output and a 5.1 bed reaches the system as two
channels, with nothing left for the OS to spatialize.

macOS pins `ao=avfoundation,coreaudio` and never had the problem. This adds the
matching override on the iOS core, `avfoundation,audiounit`, keeping
`audiounit` as the fallback. `AVSampleBufferAudioRenderer` takes a real channel
layout and its `allowedAudioSpatializationFormats` defaults to
`AVAudioSpatializationFormatMultichannel`, so spatialization needs no further
code. Guarded `#if os(iOS)`: tvOS compiles the same file, but an HDMI route
already reports more than two channels there, so `audiounit` does not clamp and
its PCM path is left alone. `MpvAudioPlayerCore` is untouched for the same
reason macOS pins only its video core.

Measured against a live mpv context, reverting the one line to produce the
before row: `current-ao` goes `audiounit` → `avfoundation` and
`audio-out-params/hr-channels` goes `stereo` → `5.1`, while
`audio-params/hr-channels` reads `5.1` in both. That last one is why the
diagnostics ship in the same change rather than after it — the performance
overlay only polled `audio-params/*`, the decoder side, so it reported the
source layout while the audio was being folded to stereo underneath it. It now
also polls `current-ao` and
`audio-out-params/{hr-channels,samplerate,format}`.

That makes this two mechanisms, and by line count the diagnostics are the
larger half. Happy to split them if you would rather they moved separately; I
kept them together because the fix is a one-line routing change with no other
way to confirm it took.

What the simulator settles is the driver selection and that `ao_avfoundation`
initialises for 6-channel PCM at all, which was genuinely unproven — it was
only reachable through `audio-spdif` before this. What it cannot settle is the
real per-route channel grant or spatialization, since its `AVAudioSession` is
backed by the host Mac's CoreAudio device; the tests therefore assert the
driver and never a channel count.

Verified: 24 `RunnerTests` cases on an iPhone 17 Pro simulator, Dart suite
green (5831 tests, 5 skipped), plus format, analyzer, codegen freshness and
translation hygiene. Both mutation directions checked — reverting the `ao`
selection fails two of the four native tests, and re-sourcing the negotiated
fields from `audio-params/*` fails four of the six Dart tests.


edde746:
Remove slop tests and unnecessary comments.

toluLikesToCode:
Closing this — the scope was wrong. Rebuilding it as the one-line driver change on its own; the diagnostics belong in their own PR if they're wanted at all.

Issue #1495, opened 2026-07-05
"Original quality" playback fails with HTTP 500 due to incorrect bandwidth calculation on Plex Media Server, while capped bitrate modes work fine


Amcro:
### Bug Description

Playback fails immediately with a "loading failed" error when quality is set to "Original," even for files well under my server's configured bandwidth limit. Switching to a fixed bitrate cap (e.g. 4mbps or 8mbps) instead of "Original" resolves the issue and playback works normally on the same file.
Comparing the client logs for both cases, the working (capped) attempts show a call to /video/:/transcode/universal/decision before playback starts, and the server responds correctly with a transcode decision. The failing "Original" attempts go straight to requesting the raw file part, with no decision call beforehand.
On the Plex Media Server side, the log for the failing request shows:
```
Streaming Resource: Bandwidth exceeded: 10212 kbps > 10000 kbps
Streaming Resource: Cannot make a decision because either the file is unplayable or the client provided bad data
Got exception from request handler: bad lexical cast: source type value could not be interpreted as target
```
The file's actual bitrate is around 5mbps, well under my 10000 kbps (10mbps) server limit, so the 10212 kbps figure the server calculated for this request doesn't match the source file. There's also a "bad lexical cast" exception logged on the same request, right after the bandwidth rejection.

### Steps to Reproduce

Steps to reproduce

1. Configure a Plex Media Server with a maximum stream bandwidth limit (in this case, 10000 kbps / 10mbps).
2. Set playback quality to "Original" in Plezy.
3. Play a file with a bitrate comfortably under the server's limit (in this case, ~5mbps).
4. Playback fails immediately with "loading failed."
5. Check Plex Media Server's log at the corresponding timestamp — it shows the bandwidth-exceeded rejection and the lexical cast exception above.
6. As a comparison, select a fixed bitrate option below the server limit (e.g. 8mbps) instead of "Original" and play the same file — playback succeeds.

### Expected Behavior

Expected behavior
Playback with "Original" quality should work for files under the server's bandwidth limit, consistent with how capped bitrate modes behave on the same files and server.

### Actual Behavior

[...]

Issue #1343, opened 2026-06-14
Server Error HTTP 500


DeepDaddyTTV:
### Bug Description

Remote users are regularly getting this error when the file is served direct play and transcoding is enabled on the server. 

### Steps to Reproduce

Play seemingly random media files from the library. 

### Expected Behavior

Media files play. 

### Actual Behavior

Playback errors out stating: Server error (HTTP 500). A bandwidth/transcoding limit likely rejected this session. Ask the owner to adjust it.

### Plezy Version

2.6.0

### Platform

Windows

### Media Server

Plex

### Device Information

Various windows computers. 

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

### Screenshots

<img width="745" height="181" alt="Image" src="https://github.com/user-attachments/assets/e92003f2-2a05-4c9d-a8fa-548f11c040f9" />

### Additional Context

I’ve tried everything from disabling transcoding on the server entirely to adjusting user settings. Nothing seems to stop it and the files it affects are seemingly random. It will occur with episode 4 but not 5 of a series with the same codec, subtitles, etc.

edde746:
Server side issue, see https://support.plex.tv/articles/227715247-server-settings-bandwidth-and-transcoding-limits/

DeepDaddyTTV:
I know that. However this issue doesn’t happen in the Plex App, my server settings follow that exact guide to avoid it, and yet it still happens randomly only in Plezy specifically on Windows.

edde746:
Plex will auto transcode opaquely for you, so you have no idea when you're getting degraded quality unless you check, Plezy does not, you have to explicitly set it to degrade the quality if that is what you want. I strongly suggest you simply disable the bandwidth limit on your server (like the error message already suggests).

This is what you need to unset:

<img width="709" height="471" alt="Image" src="https://github.com/user-attachments/assets/6511eb51-2d33-45ee-a50f-aba155d8d615" />

DeepDaddyTTV:
Yes that’s what I’m saying. It isn’t transcoding according to both Tautulli and Tracearr. It’s a direct play straight from Plex. Also, I do have the bandwidth limit disabled. Hence the confusion.

edde746:
Could you get the log from the server when it fails to play?

Issue #665, opened 2026-03-10
Fails to play random files with HTTP/500 while the official plex client works


NetSysFire:
### Bug Description

Occasionally, like right now, it just fails to play anything. The debug logs state it is an internal server error but plex web works just fine. The operator of this plex server can also not see anything wrong in his logs.

Right now I am having troubles playing a particular show but trying another one just for testing suddenly works.

I have had this before and it went away a day after. I think this one will go away after some time as well.

### Steps to Reproduce

1. Try to play anything via Plezy

### Expected Behavior

It plays.

### Actual Behavior

It does not play and reports a server error.

### Plezy Version

1.24.1.r30.g588b413-1

### Platform

Linux

### Device Information

Arch Linux x86_64

### Logs and Stack Traces

<details>
<summary>Logs for the show that is giving me a HTTP/500 (Black Mirror S3E5)</summary>

```
Plezy v1.24.1 (52)
Linux rolling
---
[12:59:42.345] [DEBUG] Navigated to home
[12:59:42.345] [DEBUG] DiscoverScreen.refresh() called
[12:59:42.345] [DEBUG] Refreshing Continue Watching in background from all servers
[12:59:42.345] [DEBUG] fetching on deck from 1 servers
[12:59:42.631] [DEBUG] fetching on deck for server 7d3f6e3589bb8fedda3700706c1bcfdf3a3bb800 completed in 285ms with 3 items
[12:59:42.631] [INFO] Fetched 3 on deck items from all servers
[12:59:42.631] [DEBUG] Continue Watching refreshed successfully
[12:59:43.751] [DEBUG] Playing: Men Against Fire
[12:59:43.764] [DEBUG] VideoPlayerScreen initialized for: Men Against Fire
[12:59:43.764] [WARNING] CompanionRemote: Cannot send command - not connected
[12:59:43.821] [INFO] [MPV:cplayer] Set property: hwdec="auto" -> 1
[12:59:43.822] [INFO] [MPV:cplayer] Set property: sub-font-size="33" -> 1
[12:59:43.823] [INFO] [MPV:cplayer] Set property: sub-color="#FFFFFF" -> 1
[12:59:43.824] [INFO] [MPV:cplayer] Set property: sub-outline-size="1" -> 1
[12:59:43.825] [INFO] [MPV:cplayer] Set property: sub-border-size="1" -> 1
[12:59:43.826] [INFO] [MPV:cplayer] Set property: sub-outline-color="#000000" -> 1
[...]

PR #1974, opened 2026-08-17
Downrank relay in the connection race and stop persisting it as preferred

cajunflavoredbob:
A relay connection can win the phase-1 race (plex.tv edge answers fast) and then gets saved as the preferred endpoint. The cached endpoint gets the head-start probe on the next launch, so one relay win pins every future session to relay's 2 Mbps cap even after direct connectivity is back. On Original quality that surfaces as the HTTP 500s from #1343 / #665. Hit this live: two remote users stuck on relay for an evening while the server's direct URL was reachable the whole time. Restarting the app didn't clear it because the relay URL was cached.

Changes:

- Non-relay candidates get an 800ms head start in the phase-1 race (new optional probeDelayOf hook on raceEndpointCandidates). If the race is decided before the handicap elapses, the relay probe is skipped entirely. Relay still wins when nothing else answers, just 800ms later.
- Relay URLs are never persisted as the preferred endpoint. Classification runs against the live registered server, not the connect-time capture, so relay URIs rotated by a profile refresh can't slip through classified as remote.
- A relay URL cached by an older version is ignored at bind, so existing installs self-heal on upgrade.

Relay stays a full fallback for people who genuinely need it (double NAT etc.); it just never becomes sticky and never beats a working direct endpoint. Doesn't touch the Original-quality playback path, so this sidesteps the transcode-API debate from #1495 entirely.

Tests cover the handicap (loses to a slower direct endpoint inside the window, still wins when alone) and all three persistence behaviors. Full suite green on 3.47.0 stable. Jellyfin discovery is untouched; it doesn't pass the hook.

### DIFF
--- a/lib/services/multi_server_manager.dart
+++ b/lib/services/multi_server_manager.dart
@@ -335,7 +335,7 @@ class MultiServerManager {
 
     // Get storage and load cached endpoint for this server
     final storage = await StorageService.getInstance();
-    final cachedEndpoint = storage.getServerEndpoint(ServerId(serverId));
+    final cachedEndpoint = _nonRelayCachedEndpoint(server, storage.getServerEndpoint(ServerId(serverId)));
 
     // The connection race already hits `/` on the winning endpoint — capture
     // `transcoderVideo` from that response so PlexClient.create can skip the
@@ -375,6 +375,13 @@ class MultiServerManager {
       serverName: server.name,
       prioritizedEndpoints: prioritizedEndpoints,
       onEndpointChanged: (newUrl) async {
+        // Classify against the live server object, not the connect-time
+        // capture: a profile refresh replaces _plexServers[serverId] with a
+        // refetched server (fresh, rotated relay URIs) while this client and
+        // closure live on, and a rotated relay URL would misclassify as
+        // remote against the stale connection list.
+        final liveServer = _plexServers[ServerId(serverId)] ?? server;
+        if (!_shouldPersistEndpoint(liveServer, newUrl)) return;
         await storage.saveServerEndpoint(ServerId(serverId), newUrl);
         appLogger.i('Updated endpoint for ${server.name} after failover: $newUrl');
       },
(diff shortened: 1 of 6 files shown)


edde746 (changes requested):
Sound design; verified locally on the repo toolchain (branch = main + 1): both touched suites pass, full `flutter test` green (5890 passed), analyzer parity clean. Two items before merge:

**Blocking: formatter.** `dart format` joins `plex_auth_service.dart:511-512` — the `probeDelayOf` closure fits on one 120-column line. This is the CI failure.

**Harden `onEndpointChanged` for the reverse rotation direction.** Classifying against the live server fixes the case the comment describes (new relay URI vs stale captured `server`), but the mirror case stays open: the client's failover list still holds connect-time URLs, the live server rotated its relay URI away, and the old relay URL — absent from the live list — falls through `_candidateForUrl` to `_classifyCustomPreferredUrl`, which classifies any non-private HTTPS host as remote, so it gets persisted, resurrecting the bug for exactly the stale-relay case. Requires the old relay URI to keep answering after rotation, so it is narrow, but the fix is one line — refuse persist when either object says relay:

```dart
if (!_shouldPersistEndpoint(liveServer, newUrl) || !_shouldPersistEndpoint(server, newUrl)) return;
```

`_promoteEndpoint` and the optimizer do not need this; their `identical(_plexServers[serverId], server)` guards make capture == live.

Notes, no change required:

- Bind-time self-heal also works when the cached relay URL has rotated, but for a different reason than classification: an unmatched preferred URL finds no candidate in `raceEndpointCandidates` (line 69) and is ignored entirely. Worth a comment on `_nonRelayCachedEndpoint` so the classifier does not get "fixed" for that case later.
- Relay-only (double-NAT) users now pay full rediscovery plus the 800ms handicap on every launch and reconnect, since relay is never cached — right tradeoff for this bug, but state it in the description; it will surface as "connects slower" reports from exactly the users relay serves.
- CI pins Flutter 3.44.0, not 3.47.0; reran on the pinned toolchain, green.

Test coverage is good: the race tests pin both handicap contracts with probe-count assertions proving the skip, and the manager tests pin all three persistence behaviors including `capturedPreferredUris == [null]`. Add a test for the reverse-rotation path only if the one-line check above lands.

edde746:
Closing in favor of https://github.com/edde746/plezy/commit/304d8a5e33500b749edbfa30b39b0a12faa5905b

PR #1999, opened 2026-08-18
fix: hide window on close

aldobarr:
After [this change](https://github.com/edde746/plezy/commit/5a25c1f9cc3cc30e18f220370f45a6497210719b) where a graceful exit is requested instead of immediately exiting, Plezy has a strange looking behavior on exit where screens disappear as they shut down and it's visible to the user. For example I click X and I see this:

Screenshot_20260818_052250

then the window itself does not close until I either press X again, or I wait a while for the window to finish closing.

To the user this looks like the app broke instead of closing like they expected. This hides the window before requesting the shutdown event so that the user experience matches their expectation of the Window just going away.

I also noticed on shutdown the app was throwing `AppLifecycleListener was used after being disposed.` so just don't emit the UI status transition since the app is closing anyway and already hidden from the user.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1280,7 +1280,7 @@ class _MainAppState extends State<MainApp> with WidgetsBindingObserver {
     await TrackerCoordinator.instance.stopPlayback().timeout(const Duration(seconds: 3), onTimeout: () {});
     TrackerCoordinator.instance.cancelInFlight();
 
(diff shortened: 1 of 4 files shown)


edde746:
Closed in favor of https://github.com/edde746/plezy/commit/436fd7d2d51157a9225d957e7335e6cdc26a1ccf

Issue #1910, opened 2026-08-13
Support Jellyfin "Enable rewatching in next up" setting


BenryBoi:
### Problem Statement

When rewatching an already-watched show on a Jellyfin server (even if episodes marked as unwatched), subsequent episodes do not automatically populate in Plezy's Continue Watching / Next Up row.

In the official Jellyfin Web client, this behavior is controlled by a client-side display setting called "Enable Rewatching in Next Up". Because this toggle is client-specific rather than global, the preference doesn't automatically carry over to Plezy.

### Proposed Solution

Add an option in the Settings/Appearance tab to display shows that are being rewatched in Continue Watching.

### Use Case

I enjoy watching shows multiple times, and it would make it so that I don't have to go scrolling through my library to find the content I was just watching, instead having it right at the top of the home page.

### Target Platform

All platforms

### Media Server

Jellyfin

### Additional Context

Plezy has been the best client I have ever used and this is honestly my only problem with it.

Ciechom:
Signed in just to say that I also use this feature extensively and until it’s implemented this app is unusable for me, I primarily watch only 1 show on repeat and have reduced mobility in my hands so juggling the library and episode list is not very user friendly. I hope dev can add this feature so I can switch fully to plezy because I find the rest of it much better then the default Jellyfin app and I also prefer it over the 3rd party competition.

PR #2086, opened 2026-08-24
feat(jellyfin): honor "Enable rewatching in Next Up" for Continue Watching

vavallee:
## What

Jellyfin web has a client-side display setting, **"Enable rewatching in next up"**, which asks `/Shows/NextUp` for `EnableRewatching` so a series the user has already finished keeps surfacing its next episode while they rewatch it. Plezy never passed the flag, so rewatched shows dropped out of Continue Watching / Next Up (#1910). Because it's a per-client setting on Jellyfin's side, nothing from the web client carries over — it needs its own toggle here.

## Change

- New appearance toggle **Rewatching in Next Up** (subtitle notes it's Jellyfin), backed by a `jellyfinRewatchingInNextUp` pref registered with the reset-and-portable set so it rides along with settings export.
- When on, `EnableRewatching=true` is sent on every `/Shows/NextUp` request the client makes: the Continue Watching shelf, the home and library hub sets, and the per-show on-deck chain (so the detail page's resume episode follows the rewatch too).
- It is only sent when the toggle is on **and** the server is Jellyfin. Jellyfin already defaults it to false, so every other request stays byte-for-byte what it was; Emby's Next Up doesn't know the parameter and never receives it (the shared on-deck chain is gated on dialect). Plex OnDeck has no equivalent switch.

API: `GET /Shows/NextUp` → `enableRewatching` (boolean, default false), "Whether to include watched episodes in next up results" (Jellyfin OpenAPI).

Strings added to every locale per the usual convention (English text, empty elsewhere).

## Tests

- `NextUp requests omit EnableRewatching while the preference is off` / `…send EnableRewatching=true once the preference is on` — capture all three NextUp request sites (shelf, hub set, on-deck) and assert the parameter's absence/presence. The "on" test fails if the flag is dropped.
- `Emby never receives EnableRewatching, even with the preference on` — Emby-dialect client, preference on, on-deck chain: no parameter. Fails without the dialect gate.
- `toggles the Jellyfin rewatching-in-Next-Up preference` (new `appearance_settings_screen_test.dart`) — the tile renders with its description, defaults off, and tapping the switch persists `jellyfin_rewatching_in_next_up`.
- `exports the Jellyfin rewatching-in-Next-Up preference` — the pref rides along with a settings export.
- `flutter analyze`, `dart format --line-length 120`, `dart_code_linter` unused-code/unused-files, `clean_translations.py --check --strict`, `check_icon_consistency`, `codegen.sh --check`, full `flutter test`: clean.

### DIFF
--- a/lib/i18n/az.i18n.json
+++ b/lib/i18n/az.i18n.json
@@ -165,6 +165,8 @@
     "visualEffectsReducedDescription": "Daha az animasiya və daha aşağı keyfiyyətli şəkillər",
     "hideSpoilers": "Baxılmamış seriyalar üçün spoylerləri gizlə",
     "hideSpoilersDescription": "Baxılmamış seriyalar üçün miniatürləri və təsvirləri bulanıqlaşdır",
+    "jellyfinRewatchingInNextUp": "",
+    "jellyfinRewatchingInNextUpDescription": "",
     "playerBackend": "Oynadıcı infrastrukturu",
     "exoPlayer": "ExoPlayer",
     "mpv": "mpv",
--- a/lib/i18n/bg.i18n.json
+++ b/lib/i18n/bg.i18n.json
@@ -165,6 +165,8 @@
     "visualEffectsReducedDescription": "По-малко анимации и изображения с по-ниска резолюция",
     "hideSpoilers": "Скривай спойлери за негледани епизоди",
     "hideSpoilersDescription": "Замазвай миниатюри и описания за негледани епизоди",
+    "jellyfinRewatchingInNextUp": "",
+    "jellyfinRewatchingInNextUpDescription": "",
     "playerBackend": "Система за възпроизвеждане",
     "exoPlayer": "ExoPlayer",
     "mpv": "mpv",
--- a/lib/i18n/da.i18n.json
+++ b/lib/i18n/da.i18n.json
@@ -165,6 +165,8 @@
     "visualEffectsReducedDescription": "Færre animationer og illustrationer i lavere opløsning",
     "hideSpoilers": "Skjul spoilere for usete episoder",
     "hideSpoilersDescription": "Slør miniaturebilleder og beskrivelser for usete episoder",
+    "jellyfinRewatchingInNextUp": "",
+    "jellyfinRewatchingInNextUpDescription": "",
     "playerBackend": "Afspillermotor",
     "exoPlayer": "ExoPlayer",
     "mpv": "mpv",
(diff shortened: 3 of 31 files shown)


edde746:
It feels really janky that preferences like this should be handled on the client. Does Jellyfin not persist this setting on the server?

vavallee:
Checked how Jellyfin itself handles this: the server doesn't persist it. There's no rewatching field in `UserConfiguration`, and `/Shows/NextUp` takes `enableRewatching` as a per-request query param (default false) — the server never applies it on its own. jellyfin-web stores the toggle client-side too: `userSettings.js` reads/writes `enableRewatchingInNextUp` with `enableOnServer: false` (localStorage, per device) and passes it on each request ([nextUp.ts#L47](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/homesections/sections/nextUp.ts#L47)). So there's nothing server-side to read — a local pref is the same design the first-party client uses.

That said, happy to rework it if the current shape isn't what you want:

- **Sync via DisplayPreferences** — stash the key in `CustomPrefs` (`/DisplayPreferences/usersettings`) so it follows the user across Plezy installs. Downside: jellyfin-web doesn't read it from there, so it still wouldn't stay in sync with the web UI, and it'd be Plezy's first use of that API.
- **Move the toggle** — if the concern is more about a Jellyfin-only switch sitting in Appearance, I can relocate it (or gate it to only show when a Jellyfin server is connected).
- Or if you'd rather not carry the toggle at all, say the word and I'll close this.

Whichever direction you prefer.

edde746:
Superseded by #2112, which takes the DisplayPreferences option from your second bullet.

Your upstream findings hold: no server-side field, `/Shows/NextUp` takes `enableRewatching` per request, and jellyfin-web passes `enableOnServer: false` so its own toggle never leaves localStorage. Where this landed differently is the conclusion drawn from that. `CustomPrefs` is keyed `(userId, displayPreferencesId, client)` with no device component, so it does give a per-account answer even though jellyfin-web declines to use it — that jellyfin-web keeps the switch local is a fact about jellyfin-web, not a constraint on us. Verified on 10.11.10: written from one device and token, read back from another, and the resulting `/Shows/NextUp?…&EnableRewatching=true` returns 200.

Two things that came out of implementing it. The row has to be Plezy's own `client` name, not the `emby` one that jellyfin-web writes and jellyfin-androidtv reads: a POST there replaces the whole custom set and resets any known scalar the body omits — `skipForwardLength` went 15000 → 30000 in testing — so sharing it would trash their skip lengths and home-section order on every save. And Emby needs gating out entirely, since its `/Shows/NextUp` has no `EnableRewatching`; that is now `MediaBrowserDialect.supportsNextUpRewatching`.

Your request plumbing carried over as-is: the same four call sites (Continue Watching shelf, home and library hub sets, per-show on-deck chain) and the same dialect gate. What changed is where the value comes from — an account row instead of a local pref — and where the toggle lives: a new Account preferences settings section that also covers the `UserConfiguration` fields and the plex.tv account ones, rather than a Jellyfin-only switch in Appearance.

Closing this; #1910 closes with #2112.

PR #2091, opened 2026-08-24
Add Emby remote subtitle search and download flow

Bockiii:
## Summary

Adds Emby subtitle search/download support and wires it through the existing subtitle controls flow.

This also generalizes the previous Plex-only subtitle search path into a backend-neutral model/API so Plex and Emby can use the same UI and apply pipeline.

## What changed

- added Emby remote subtitle search via `/Items/{id}/RemoteSearch/Subtitles/{language}`
- added Emby subtitle download support via the remote subtitle download endpoint
- added post-download apply handling for Emby, including use of `NewIndex` when the server returns it
- generalized subtitle search result handling from Plex-specific to shared subtitle DTOs
- extended the media client interface with shared subtitle search/download/source-track hooks
- updated Plex to use the shared subtitle search abstractions
- gated subtitle search visibility for Emby using user policy flags
- updated track controls/search sheet flow to work against backend capabilities instead of Plex-only assumptions

## Notes

- `TEMPFORDEV` and debugging-only instrumentation were removed and are not part of this change.
- The remaining diff is only product code for the subtitle feature.
- The Emby capability gate now follows the explicit subtitle policy flags instead of Premiere heuristics.

## Validation

- verified Linux build/debug runs during implementation
- validated touched files with `flutter analyze`
- manually tested Emby subtitle search and download flow
- confirmed downloaded subtitles appear in Emby and in Plezy
- confirmed subtitle search gating behavior follows current server policy responses

## Follow-up / caveat

The visibility gate now intentionally trusts Emby’s subtitle policy flags (`EnableSubtitleDownloading` / `EnableSubtitleManagement`). If a non-Premiere server reports those as enabled, Plezy will expose subtitle search accordingly because no stronger Premiere-specific signal was found.

### DIFF
--- a/lib/media/media_server_client.dart
+++ b/lib/media/media_server_client.dart
@@ -7,6 +7,7 @@ import '../utils/app_logger.dart';
 import '../utils/media_server_http_client.dart' show AbortController, MediaServerResponse, throwIfHttpError;
 import '../utils/external_ids.dart';
 import '../utils/watch_state_notifier.dart';
+import '../models/subtitles/subtitle_search_result.dart';
 import 'artist_discography.dart';
 import 'download_resolution.dart';
 import 'ids.dart';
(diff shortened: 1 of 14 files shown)


Bockiii:
Well... this PR started yesterday, before you reworked the plex subtitle handling and took care of the caching/selection error that I have also fixed with this. Not sure what to do now.

Bockiii:
I am reverting all your changes in 8b5ee4f as i've already implemented that fix in my unified solution

Bockiii:
Alright, the plex-fix of yours threw a little spanner in the works, but it should all work now. I've tested this in the linux and the windows client.

[emby_subtitles.webm](https://github.com/user-attachments/assets/d3be4e34-b010-41a7-9894-26d80dc644d6)
[plex_subtitles.webm](https://github.com/user-attachments/assets/52d7a1eb-7161-4328-822b-50555bafc33b)

This is a test on linux client with plex and emby (with premiere).

After squashing the chaotic commits, it should be good to go :)

Bockiii:
FYI, I expect the emby guys to probably just send you a lifetime premier key if you send them a message, saying you're the plezy dev.

Bockiii:
The longer you wait for merging, the more follow up issues might occur if you change/fix something else with subtitles

PR #2094, opened 2026-08-24
feat(player): enable clickVideoTogglesPlayback by default on desktop

Squiddim:
enables click to play/pause on Desktop Clients by default

For the first months of using Plezy I really missed being able to just click on the video to play/pause. I only just found out that this is implemented but disabled by default.
IMO such an intuitive control should be enabled by default, as it is how users interact with other players too.


Squiddim:
ok not even a comment thx I guess ...

edde746:
This is just a disagreement of what the default behavior should be.

PR #2108, opened 2026-08-24
feat(seerr): sign in with Jellyfin Quick Connect and accept bare-host URLs

l3gitpanda:
## What

Two fixes to linking a Seerr instance, both on the Jellyfin path:

**Quick Connect sign-in.** The Jellyfin sign-in form gains a "Use Quick Connect" button that shows the same full-screen code panel as the Jellyfin/Emby add-server flow — large code, "waiting for approval", cancel — and finishes the link once the code is approved inside Jellyfin. Until now the only way to link Seerr with a Jellyfin account was typing a username and password, which on a TV remote is miserable.

**Forgiving URL entry.** The URL step prepended `https://` to anything without a scheme, so a plain-HTTP instance on the LAN failed with "could not reach" unless the user knew to type the scheme — `192.168.1.5:5055` and bare hosts on Seerr's default port both fell into this. Schemeless input is now expanded into candidates and raced, the same way the MediaBrowser add-server form already treats server URLs.

## Screenshots

image

## How

- `SeerrAuthService.expandUrlCandidates` (pure, static, unit-tested) mirrors `JellyfinEndpointDiscovery`'s rules: an explicit scheme is authoritative, `host:port` tries `https://` then `http://`, and a bare host tries `https://host`, `http://host`, `http://host:5055` (Seerr's default port). Sub-paths are preserved on every candidate. `probeFirstReachable` probes the candidates concurrently and takes the first that answers `/settings/public` as an initialized instance; when none does it rethrows the *first* candidate's error so the message names the URL the user meant. Guesses stay discovery-only — only the winning URL reaches the session.
- Quick Connect goes through Seerr's own proxy routes (added in Seerr 3.4.0, seerr-team/seerr#2212): `POST /auth/jellyfin/quickconnect/initiate` → `{code, secret}`, then `GET …/check?secret=` polled through the shared `pollWithBackoff` at Seerr's own 2s cadence — a 404 mid-poll means the secret expired and is terminal — then `POST …/authenticate`, which returns the user and `connect.sid` through the existing `_signIn` path. Attempt-id and cancel guarding mirror `AddJellyfinScreen` so a dismissed screen can't setState late.
- New `SeerrAuthMethod.quickConnect`. Sessions serialize the method by name, so appending is safe, and this method deliberately stores no secret: silent re-auth is impossible by construction, so an expired cookie lands in the existing "no stored credentials" arm and unlinks the session, prompting a fresh code instead of retrying forever.
- The button is limited to the Jellyfin credential form — Seerr rejects Quick Connect for Emby. It is deliberately **not** auto-started on TV (unlike `AddJellyfinScreen`, which auto-jumps): Seerr exposes no "Quick Connect enabled" capability in `/settings/public`, so auto-firing would blind-hit instances that can't serve it. On a pre-3.4 instance the initiate call 404s and the error surfaces inline under the form. Happy to switch to an explicit version check if you'd rather gate it that way.
- No new i18n keys: the panel reuses `auth.useQuickConnect` / `auth.quickConnect*` and the `addServer.quickConnect*` error strings. "Quick Connect" is added to the hardcoded-strings allowlist for the settings-screen method label.

## Testing

- 10 new unit tests in `test/services/seerr/seerr_client_test.dart` covering candidate expansion (scheme, bare host, explicit port, sub-path, blank), the http/default-port fallback race and the all-candidates-fail error, initiate, poll-then-authenticate, secret expiry mid-poll, and that a `quickConnect` session refuses silent re-auth.
- Manually verified on Windows desktop against a live Seerr instance backed by Jellyfin: Quick Connect sign-in completes end to end after approving the code in Jellyfin.
- `flutter analyze` clean, `dart format` (page width 120) applied, `scripts/checks/clean_translations.py --check --strict` and `scripts/checks/check_hardcoded_strings.py` pass, and the full `flutter test` suite is green apart from failures that reproduce identically on an unmodified checkout of this base.

AI-assisted: implemented with Claude Fable 5 via Claude Code.

### DIFF
--- a/lib/models/seerr/seerr_session.dart
+++ b/lib/models/seerr/seerr_session.dart
@@ -15,6 +15,11 @@ enum SeerrAuthMethod {
 
   /// `POST /auth/local` with stored email/password.
   local,
+
+  /// `POST /auth/jellyfin/quickconnect/authenticate` (Seerr 3.4+). No stored
+  /// secret — silent re-auth is impossible, so an expired cookie unlinks the
+  /// session and the user reconnects with a fresh code.
+  quickConnect,
 }
 
 /// Which Seerr product the instance runs. The two products disagree on
--- a/lib/screens/settings/seerr_connect_screen.dart
+++ b/lib/screens/settings/seerr_connect_screen.dart
@@ -9,6 +9,7 @@ import '../../mixins/controller_disposer_mixin.dart';
 import '../../models/seerr/seerr_public_settings.dart';
 import '../../models/seerr/seerr_session.dart';
 import '../../providers/seerr_account_provider.dart';
+import '../../services/seerr/seerr_auth_service.dart';
 import '../../services/seerr/seerr_constants.dart';
 import '../../services/seerr/seerr_exceptions.dart';
 import '../../theme/mono_tokens.dart';
@@ -21,10 +22,13 @@ import 'async_form_state_mixin.dart';
 enum _CredentialForm { none, jellyfin, emby, local }
 
 /// Two-step Seerr connect flow:
-///   1. Probe the instance URL (`/settings/public`).
+///   1. Probe the instance URL (`/settings/public`), expanding schemeless
+///      input into https/http/default-port candidates like the MediaBrowser
+///      add-server form.
 ///   2. Sign in with one of the methods the instance supports — one-tap
 ///      Plex (reusing the profile's stored token), Jellyfin/Emby
-///      credentials, or a local Seerr account.
+///      credentials, Jellyfin Quick Connect (Seerr 3.4+), or a local Seerr
+///      account.
 ///
 /// The finished [SeerrSession] is handed to [SeerrAccountProvider.adoptSession]
 /// and the screen pops.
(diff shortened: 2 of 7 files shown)


edde746:
Both behaviors are on main, implemented differently: `75279ca1` (shared URL-candidate expander) and `595cd9fe` (Seerr Quick Connect + schemeless URL entry). Ships in the next release.

Carried over from here: the proxy-route sequence (`initiate` → `check` at Seerr's 2s cadence → `authenticate`), the terminal 404 mid-poll, a `quickConnect` method that deliberately stores no secret so an expired cookie unlinks instead of retrying forever, the Jellyfin-only gate, and no auto-start on TV.

Why it wasn't merged as-is:

1. **The candidate race could hand a password to plaintext.** `probeFirstReachable` raced https and http together, so when a host serves Seerr over both, the http probe wins whenever the TLS handshake is slower — reproduced with a 120ms TLS delay, winner `http://seerr.example.com`. That URL is what `signInWithJellyfin`/`signInWithLocal` then posts to, and what the session persists. On main a plaintext success is held while any https candidate is still in flight and is accepted only once they have all failed, bounded by the existing probe timeout.
2. **Reporting the first candidate's error masked the actionable one.** An uninitialized instance on plain http with https refused reports `Could not reach https://seerr.example.com`, not "has not completed first-run setup". `SeerrUrlException` now carries the status of the response that disqualified a URL, and a response-derived failure outranks a transport error.
3. **`Uri.replace(query: null, fragment: null)` is a no-op** — null means "keep mine", so `example.com/seerr?a=1#top` produced base URLs still carrying the query that request paths get appended to. Candidates are built field by field now.
4. **The duplication.** `expandUrlCandidates` was `JellyfinEndpointDiscovery.expandInputToBaseUrls` with the port swapped, plus a third copy of the scheme regex; `_buildQuickConnectPanel` was byte-identical to `add_jellyfin_screen.dart`'s modulo enum shorthand, and `_cancelQuickConnect`/`_requestFocusAfterFrame` identical. Main has one `expandBaseUrlCandidates(input, guesses:)` in `url_utils.dart` — each backend passes its own ordered guesses, and a typed port beats a guessed one so bare hosts and `host:port` share one list — plus one `QuickConnectCodePanel` and one `QuickConnectFlowMixin` used by both screens.
5. **The Quick Connect secret was not registered for redaction.** It travels in a query string and `ClientException.toString()` includes the URI, so a poll blip could log it. `LogRedactionManager.registerCustomValue` now covers it, as on the Jellyfin path.
6. **Strings.** "Quick Connect" went into the global brand-name allowlist; it is `auth.quickConnect` now. A pre-3.4 instance gets "This Seerr instance does not support Quick Connect. It needs Seerr 3.4 or newer." instead of "Quick Connect failed: HTTP 404".

Coverage: 12 service cases (TLS beats fast plaintext, response error preferred over transport, hostless input probes nothing, initiate 404, poll→exchange, secret expiry, cancel without exchange, no silent re-auth), 6 expander cases, and 7 widget cases driving the connect screen end to end.

l3gitpanda:
Sounds good! Thanks for maintaining the project and doing such great work! Would appreciate it if you could look into a few of my other PRs. I'm not sure if you want to merge them as is or implement something similar but they would both be major quality of life features for my users. 

feat(seerr): request missing movies and seasons from library items
https://github.com/edde746/plezy/pull/1523

feat(livetv): add guide search with jump-to-channel/program
https://github.com/edde746/plezy/pull/1526

fix(player): fullscreen button icon stuck on "enter fullscreen" on Windows
https://github.com/edde746/plezy/pull/1531

PR #2145, opened 2026-08-27
fix(player): preserve packed aspect for stereoscopic (3D) video

RyanTheTechMan:
Fix an issue that causes standards-tagged Matroska Side-by-Side and Over-Under 3D video to use per-eye geometry for the complete packed frame, making the picture appear squeezed.

## Changes
- Detect Matroska StereoMode in mpv ~~and Android ExoPlayer~~, including the Media3 and FFmpeg demux paths.
- Preserve the complete packed SBS or Over-Under frame without splitting, converting, or re-encoding the video.
- Preserve left/right eye ordering and use the correct packed-frame aspect.
- Keep ordinary 2D, custom aspect ratio, and anamorphic playback behavior unchanged.
- Restore normal sizing when switching back to a 2D source.

This goes hand in hand with #2146, which makes Plezy's playback controls usable while the display is in its packed 3D mode.

Tested on an NVIDIA Shield using tagged SBS content mpv.

AI-assisted using OpenAI GPT-5.6 Sol.

### DIFF
--- a/lib/services/video_filter_manager.dart
+++ b/lib/services/video_filter_manager.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


edde746:
The plan is to remove ExoPlayer soon, could you remove the ExoPlayer parts and rebase on main?

RyanTheTechMan:
Done. ExoPlayer 3d changes have been removed. Moved to it's own branch on my fork

PR #2165, opened 2026-08-28
feat(explore): open matched items from explore, in library

### LINKED ISSUE #1917: [Feature Request] Add a Play button for Watchlisted items that are locally available
### Problem Statement

Current Plezy version: 2.13.0

Issue: 
When a Watchlisted item becomes available to play, you have to scroll to go past a very long and frankly unnecessary amount of information to find your server. THEN you click on it, and THEN you click play. 
That's a lot of scrolling and clicking just to play the content a user has Watchlisted. 

Original Screenshot:

Image

### Proposed Solution

Proposed Changes: 

Image

The only difference is to bring the play icon (and optionally the download button) right up there so that the user can just hit play and start watching the thing they want quickly.

### Alternatives Considered

_No response_

### Use Case

1. This would save a lot of time for people who watchlist a lot of content, but then forget about the stuff they watchlisted. 

2. This can potentially also help Plezy users that are on different platforms like TVs where minimizing needless scrolling and clicking becomes important

### Target Platform

All platforms

### Media Server

Plex

### Additional Context

_No response_

RyanTheTechMan:
Closes #1917 

## Summary
Adds an “Open in library” action to Explore catalog item details when Plezy finds matching items on connected media servers.

This reuses the existing `CatalogLibraryMatcher` results rather than introducing another lookup path.

## Behavior
- The action stays hidden while matches are loading or when no matches exist.
- One match opens its normal server-specific media detail page directly.
- Multiple matches open a chooser containing every individual media copy.
- Chooser entries show the server name with quality information underneath.
- No Play or Download action is added. (issue was directly asking for this feature, but that doesn't fit how media clients work)

Used GPT 5.6 Sol to help with development.

https://github.com/user-attachments/assets/bbd11a59-c919-4e61-9d1e-37b07c057d66

---

Note:
I left the "In these libraries" section alone. In theory it *could* be removed.
CleanShot 2026-08-28 at 04 00 57

### DIFF
--- a/lib/i18n/az.i18n.json
+++ b/lib/i18n/az.i18n.json
@@ -1241,6 +1241,7 @@
     "removedFromWatchlist": "Baxış siyahısından silindi",
     "watchlistUpdateFailed": "İzləmə siyahısı yenilənə bilmədi",
     "watchlistNoMatch": "Bu elementi heç bir baxış siyahısı ilə uyğunlaşdırmaq olmadı",
+    "openInLibrary": "Kitabxanada aç",
     "notInLibrary": "Kitabxananızda yoxdur",
     "inTheseLibraries": "Bu kitabxanalarda var",
     "checkingLibrary": "Kitabxananız yoxlanılır...",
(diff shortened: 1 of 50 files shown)


edde746:
This just seems to duplicate the "In these libraries", what's the point?

RyanTheTechMan:
The main benefit is on mobile, where “In these libraries” is farther down the page. This PR puts the shortcut alongside the other actions, like in the real Plex mobile app, so users don’t have to scroll through the whole page first to get to a common action.

I kept the lower “In these libraries” because it’s familiar to desktop users (coming from Plex desktop), but I agree it’s redundant and it’s up to you if removing it makes more sense, but that may break expected workflow on desktops/larger screens.

Issue #791, opened 2026-04-02
TV Theme Music (Support Local Media Assets)


JunaidDawood:
### Problem Statement

Please add support for https://support.plex.tv/articles/201448016-how-do-i-add-theme-music-for-a-tv-show/

### Proposed Solution

Support the local media asset metadata agent source for own MP3 files

### Target Platform

Windows, All platforms

MaelRivera:
Agree.

Please, add Theme Music from Local Assets.

Not only for TV Shows, but for Movies too. The way the current Plex Apple TV app and the Plex Web app work now.

Many Thanks!!!

kvart714:
I opened a PR that implements Plex theme music support:

https://github.com/edde746/plezy/pull/2170

It adds theme music playback on media details, TV spotlights, and the mobile home hero. It uses Plex's theme endpoint, coordinates with existing music/video playback, and includes a user setting to limit playback to detail screens or enable it on the home screen as well.

I would appreciate feedback on whether this matches the intended scope for local theme media assets before closing this issue.

PR #2170, opened 2026-08-29
feat(plex): play theme music on home spotlights

kvart714:
Related to #791

## Summary
- Play Plex theme music on TV spotlights, mobile home heroes, and media details.
- Coordinate theme audio with music and video playback, including fade transitions.
- Add an appearance setting and translation updates for all locales.

## Validation
- `flutter test test/services/playback_coordinator_test.dart`
- `flutter analyze` (no issues)
- Translation hygiene and icon consistency checks
- CI guard scripts
- Android debug Docker build: `arm64-v8a` and `armeabi-v7a`

## Known Base-Branch Failures
- `scripts/codegen.sh --check` reports stale `server/relay_protocol_gen.go` already on `origin/main`.
- `scripts/run_tests.sh` fails in existing `discover_screen_test.dart` because the unchanged `TvSpotlightController` leaves a `Debouncer` timer pending.

AI-assisted: GitHub Copilot


edde746:
> GitHub Copilot

Which model?

This should also absolutely not be enabled by default.

JunaidDawood:
Does this also support movie themes?

kvart714:
> Does this also support movie themes?

I use the standard Plex API method for theme music. As far as I know, Plex provides theme music out of the box only for TV shows, but not for movies.

However, I just tried putting a `theme.mp3` file next to a movie file, and it started playing on the movie page! So Plex supports local metadata, including music, and provides it through the same API method.

As far as I know, there are also ready-made plugins that can download  `theme.mp3`  files for the whole media library

kvart714:
> Which model?

All of them 😉 But I’m not sure which model generated the PR description

> This should also absolutely not be enabled by default.

In Plex, this is enabled by default, so when I first opened Plezy, my first impression was: “Wow, but where is the theme music?!” Everything looks better, but you notice the missing music right away. So I decided to spend my weekend adding it 😄 

I agree that music definitely should not be enabled by default on the home screen. But I think having music only on the Details screen by default works well. What do you think?

But I’m also ready to turn it off completely by default if you think that would be better.

kvart714:
@edde746 Hello! I’ve been using the debug version on my TV for about a week, and I think it looks and sounds great 😍

Could you please tell me if I can expect this feature to be merged? Do I need to resolve the conflicts every time, or would it be better to build the APK I need and stay on my own fork?

eas125:
> @edde746 Hello! I’ve been using the debug version on my TV for about a week, and I think it looks and sounds great 😍
> 
> Could you please tell me if I can expect this feature to be merged? Do I need to resolve the conflicts every time, or would it be better to build the APK I need and stay on my own fork?

+1 Hopefully gets merged. I just purchased recently and it's one of the features I noticed and missed having.

JunaidDawood:
Yes, please add this to the next release since the previous Plex experience (https://support.plex.tv/articles/201448016-how-do-i-add-theme-music-for-a-tv-show/) supported both TV and movie themes. Thank you!

NeuronButter:
> I agree that music definitely should not be enabled by default on the home screen. But I think having music only on the Details screen by default works well. What do you think?

FWIW I'm keen on this feature, and do agree with this. The official Plex clients don't play music on the home screen (seems like it'd be a nuisance anyway), but the details page should have it. Movies have also been supported for a while so it'd be natural to have this implemented as well, server-side they just require a local asset since they won't fallback on Plex's default 30-second snippet service.

JunaidDawood:
It looks like Plex officially brought back this feature in their beta for the new experience:

https://forums.plex.tv/t/it-s-been-a-year-where-is-theme-music-on-the-mobile-app-when-is-music-coming-back-to-app-etc/938927/4

https://old.reddit.com/r/PleX/comments/1welu2d/looks_like_theyve_brought_tv_theme_music_back/

Issue #2034, opened 2026-08-19
Nvidia Shield Pro - After Plezy goes to sleep it remains stuck in Profile Picker


DJScias:
### Bug Description

Once the Shield goes into sleep mode during a video and you resume the video stops and you are redirected to the "Choose Profile" screen.

But you are not able to choose a profile and have to force stop the app to fix. (Probably due to "Ask for profile on app open".)

### Steps to Reproduce

1. Turn on "Ask for profile on app open".
2. Open a stream.
3. Let Shield go into sleep mode (screen off).
4. Wake Shield up and be presented the "Choose Profile" screen.
5. Notice you can't choose a profile or continue, you are stuck.

### Expected Behavior

You can choose the profile you were on last so you can resume the video stream.

### Actual Behavior

You are stuck on the "Choose Profile" screen without being able to choose anything and have to force close the app.

### Plezy Version

2.15.0

### Platform

Android TV

### Media Server

Plex

### Device Information

Nvidia Shield Pro (2019)

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

Issue #1990, opened 2026-08-17
Fire TV Cube: Alexa invocation returns Plezy to profile selection


EarzFish:
### Bug Description

On a Fire TV Cube (3rd generation), invoking Alexa while Plezy is open causes Plezy to return to the profile selection screen once the Alexa interaction finishes.

This happens even when the Alexa command has absolutely nothing to do with Plezy, Plex, playback, or the television.

For example:

“Alexa, turn the lights off.”

Alexa correctly turns the lights off, but when the Alexa overlay closes, Plezy has returned to the profile selection screen instead of remaining where it was.

### Steps to Reproduce

1. Open Plezy on a Fire TV Cube.
2. Select a Plex profile.
3. Navigate normally within Plezy / start using the app.
4. Invoke Alexa using the Fire TV Cube.
5. Give Alexa an unrelated smart-home command, e.g.:
6. “Alexa, turn the lights off.”
7. Allow Alexa to complete the command and close its overlay.

### Expected Behavior

Plezy should resume exactly where it was before Alexa was invoked.

Invoking the Fire TV Cube's Alexa interface should not cause Plezy to restart its profile-selection flow.

### Actual Behavior

Plezy returns to the profile selection screen. Playing file remains waiting behind profile selection.

The previously selected profile effectively needs to be selected again.

### Plezy Version

2.13.0

### Platform

Android TV

### Media Server

Plex

### Device Information

Fire TV Cube 3rd Gen

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

### Additional Context

Device: Amazon Fire TV Cube, 3rd generation

Plezy version: 2.13.0

The Alexa command does not need to relate to playback or the TV.

This appears to happen simply because invoking Alexa temporarily interrupts/backgrounds Plezy.

It may therefore be related to how Plezy handles the Android/Fire OS app lifecycle when the application temporarily loses focus and is resumed.

edde746:
Logs?

EarzFish:
Ah. Now I need to figure out how to get logs off my cube. I'll adb later and see if I can pull that way.

edde746:
Settings > View Logs and press the upload button, also enable debug logging in settings before reproducing the issue.

EarzFish:
LogID kgqgt

I turned opened app, tested upload logs, turned on verbose, called alexa and signed back in after the prompt came back. Tried to upload logs but it failed so I had to exit app and upload again.

[...]

Issue #2087, opened 2026-08-24
Android - Switching apps & Device Sleep immediately requires profile picker & pin auth again


DJScias:
### Bug Description

Following fixing of: https://github.com/edde746/plezy/issues/2035

I was testing out companion remote, which means I would have the device switch apps or go to sleep.

However immediately after your device goes to sleep or you switch the app and you then re-open Plezy.. It will present the user with the "profile picker" menu and have to put in the pin again.

### Steps to Reproduce

1. Open Plezy.
2. Enable Companion Remote (optional).
3. Switch to another app / let device to go to sleep.
4. See you have to re-choose profile and re-enter pin.

### Expected Behavior

1. Open Plezy.
2. Enable Companion Remote (optional).
3. Switch to another app / let device to go to sleep.
4. As long as app was not killed, let you instantly resume what you were doing.

### Actual Behavior

You are put on the profile picker and have to input the pin again immediately after app switching or device sleep/lock.

### Plezy Version

2.17.0

### Platform

Android

### Media Server

Plex

### Device Information

Oneplus 12 (OxygenOS 16)

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

### Additional Context

While very good security, I feel if the user has authenticated and just switches to another app / has device go to sleep the profile picker should be skipped.

Of course if the user kills the app then it seems logical to go through the profile picker and PIN flow again.

edde746:
Do you have the option for it enabled?

DJScias:
> Do you have the option for it enabled?

Yes, I have the option enabled.

However, given what Plezy offers, there is a specific interaction that feels unintended / counter-productive: Companion Remote.

Plex has the same option and does close on app sleep or switching out. However, Plex's app usage is strictly "actual local use". 
It does not feature a "background idea" like Companion Remote.

If completely emulating this behavior like the Plex app is intended, then I suppose this report makes little sense.

However, it feels logical that since Plezy has the Companion Remote, being connected to something should keep that session active.
Nobody keeps the app continuously on-screen without sleep or switching apps at times.

Apologies again if this is the intended use-case, but it felt off when you used it in practice so I figured I'd report it.

DJScias:
Thank you very much for considering and implementing this feature request.

I'll be sure to test it out once the next version releases.

DJScias:
Can confirm it all works fine as of 2.17.1.

PR #2195, opened 2026-08-31
fix(profiles): show the resume profile picker over active playback

markwellis:
"Ask for profile on app open" used to raise the picker on resume; #2034 stopped it whenever a video player was open, by adding a blanket resume-time exemption for active playback. That is the case the setting exists for — a show left paused overnight lets anyone resume the session on a PIN-protected profile.

Drops the `hasActiveVideoPlayback` exemption from `shouldShowProfileSelectionOnResume`. The picker still only pushes when `requiresSelectionOnOpen` holds (setting on, ≥2 profiles), so users without the setting are unaffected. #2034's focus fix (`isRouteChainCurrent`) is independent and stays, so the picker is responsive over the player.

Verified on a Fire TV Cube: with the setting on and a show playing, "Alexa, go home" → reopen shows the picker; an Alexa overlay ("what's the weather") still does not (#1990); `flutter test` green (6435 pass).

### DIFF
--- a/lib/screens/main_screen.dart
+++ b/lib/screens/main_screen.dart
@@ -98,28 +98,24 @@ bool shouldHandleDesktopRootEscape({
 ///
 /// Mobile-only: on desktop `resumed` fires on every window focus gain
 /// (alt-tab, click), which is far too frequent — the startup prompt is
-/// sufficient there. Never during active video playback: waking the device
-/// mid-stream must resume the stream, not stack the root-navigator picker
-/// over the live player route, whose focus self-heal fights the picker for
-/// the remote (#2034) — the playback session already belongs to the profile
-/// that started it. Likewise never during a live companion-remote session:
-/// a phone driving another device backgrounds and sleeps constantly, and
-/// the picker + PIN would bury a session that already belongs to the
-/// profile that started it (#2087).
+/// sufficient there. The picker still shows over an active player: the user
+/// enabled "ask on open" precisely so a different person cannot resume their
+/// session, and #2034's focus fix keeps it responsive. Never during a live
+/// companion-remote session: a phone driving another device backgrounds and
+/// sleeps constantly, and the picker + PIN would bury a session that already
+/// belongs to the profile that started it (#2087).
 @visibleForTesting
 bool shouldShowProfileSelectionOnResume({
   required bool resumedFromBackground,
   required bool isOffline,
   required bool alreadyShowingProfileSelection,
   required bool isMobilePlatform,
-  required bool hasActiveVideoPlayback,
   required bool hasActiveCompanionRemoteSession,
 }) {
   return resumedFromBackground &&
       !isOffline &&
       !alreadyShowingProfileSelection &&
       isMobilePlatform &&
-      !hasActiveVideoPlayback &&
       !hasActiveCompanionRemoteSession;
 }
 
@@ -1151,7 +1147,6 @@ class _MainScreenState extends State<MainScreen>
       isOffline: _isOffline,
       alreadyShowingProfileSelection: _isShowingProfileSelection,
       isMobilePlatform: Platform.isAndroid || Platform.isIOS,
-      hasActiveVideoPlayback: VideoPlayerScreenState.activeGlobalKey != null,
       // Short-circuit on resumedFromBackground: the provider is lazy and
       // otherwise unused on phones, so an unconditional read would create it
       // on the first lifecycle event for users who never open the remote.
--- a/lib/screens/main_screen.dart
+++ b/lib/screens/main_screen.dart
@@ -140,7 +140,11 @@ class ProfileSelectionResumeGate {
   /// Feeds one lifecycle transition through the gate. Returns true exactly
   /// once per backgrounding: on the first `resumed` after the sequence
   /// reached `hidden`/`paused`/`detached`. Consuming resets the latch.
-  bool consumePromptOn(AppLifecycleState state) {
+  ///
+  /// `pipContinuation` (read at the backgrounding, not at resume, since PiP can
+  /// end before the resume arrives): the session keeps playing in Picture in
+  /// Picture, so it never left the user's sight and must not prompt (#2195).
+  bool consumePromptOn(AppLifecycleState state, {bool pipContinuation = false}) {
     switch (state) {
       case AppLifecycleState.resumed:
         final shouldPrompt = _wasBackgrounded;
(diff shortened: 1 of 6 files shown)


markwellis:
Note, this also applies for when I turn my tv off mid episode, which is a real problem when the kids start plezy the next morning and it resumes showing whatever I was watching the night before

edde746 (changes requested):
Direction is right — the exemption defeated the setting, and #2034's focus fix stands on its own: `isRouteChainCurrent` walks ancestor navigators and the root-picker-over-nested-player shape is covered in `test/utils/route_visibility_test.dart`, so the wedged picker can't return. Two gaps before this merges:

**1. Playback auto-resumes behind the picker.** With a stream *playing* at sleep (the #2034 repro), `_handleAppHidden` pauses and latches `_wasPlayingBeforeInactive` (`lib/screens/video_player/parts/lifecycle.dart:112`), and on `resumed` the player's `restoreAfterResume` (`lib/screens/video_player/media_controls_screen_controller.dart:156-167`) unconditionally seeks back and plays — the Android TV suspend/restore path lands in the same call. Result: opaque picker on top, audio playing and the server session progressing underneath before anyone entered a PIN. The session resumes itself; the picker only hides it. Gate the auto-play on `isRouteChainCurrent` so a covered player stays paused — that is the protective behavior this PR argues for, and the uncovered resume path is unchanged. The Fire TV test in the description ("a show playing, go home, reopen") should already reproduce this as audio behind the picker.

**2. iOS auto-PiP now trips the prompt.** With `autoPip` on iOS, pressing home mid-play backgrounds the app while PiP keeps playing (`_shouldSkipForPip` skips the pause, `video_player_screen.dart:1099`); Flutter delivers `paused`, which latches `ProfileSelectionResumeGate`. Returning to the app then stacks the picker over a session that never left the user's sight — the same continuity argument that keeps the companion-remote exemption (#2087). The deleted check was implicitly the PiP guard. Don't query `PipService().isPipActive` at `resumed` time (PiP can end before the event arrives); latch the decision where the gate latches — at `paused` the player already knows it skipped the background pause for PiP — and consume it on resume. Android auto-PiP is unaffected as-is: PiP holds the activity at `inactive`, which never latches the gate.

Verified fine, no change needed: picking a *different* profile over a live player tears down cleanly — the `ProfileSessionScreen` remount disposes the player route, and `dispose()` handles externally removed routes (`_sendStoppedProgressOnce`, tracker stop, `_activeRouteGuard.clear`).

markwellis:
Both addressed:

**1. Covered player stays paused.** `restoreAfterResume` gates the auto-play on `isRouteChainCurrent`, so a picker-covered player no longer seeks+plays under it; the uncovered resume path is unchanged. Verified on a Fire TV Cube — show playing, go-home → reopen shows the picker with the stream silent underneath.

**2. iOS auto-PiP.** Guard reinstated, latched at `paused` not queried at resume: the gate skips arming when the active player reports `_shouldSkipForPip` (`VideoPlayerScreenState.activePlayerContinuesInPip`). Android is unaffected — it holds at `inactive`, which never latches.

`flutter test` green (6437).

markwellis:
I've been running this build for 2 days, works so far. But I had once an issue where the profile selector wasn't focused (couldn't change or select anything with remote) but I could play/pause the video (only got audio). I've added a new commit to address that :crossed_fingers:

edde746 (changes requested):
Commits 1–2 resolve both points. Commit 3 doesn't hold: the focus wedge you hit isn't an autofocus race at push time, and the one-shot re-assert in `ProfileSwitchScreen` can't beat where the steal actually comes from.

**The wedge is `_claimPlayerSurfaceFocus`.** Autofocus applies whenever the picker's route scope has no focused child — always true for a fresh route — so the tile wins the first frame. Focus is taken *later*, by the controls' own claims, none of which check the route chain:

- `lib/widgets/video_controls/parts/visibility.dart:322-332`: `_claimPlayerSurfaceFocus` calls `_focusNode.requestFocus()` immediately with only a sheet check, and its post-frame retry (line 329) uses the naive `ModalRoute.of(context)?.isCurrent` — the same #2034 shape, since the nested player route stays current under a root picker.
- It fires from `_onChromeChanged` on every chrome hide (`visibility.dart:280`). `setPlaying(false)` starts `startPausedAutoHide` (`player_chrome_controller.dart:61`) and `setHasFirstFrame(true)` restarts auto-hide (`:72`); the Android TV suspend-restore path goes through `_firstFrame.resetUiForOpen()` (`playback_reload.dart:500`) and re-arms it seconds after resume. A paused, picker-covered player still hides its chrome on a timer and pulls focus off the picker — which matches what you saw (picker unfocused, D-pad select play/pausing audio underneath).
- `_requestPlayPauseFocus` (`visibility.dart:336`) has no route check at all.

Drop `_assertInitialFocusOnce` and gate `_claimPlayerSurfaceFocus` (immediate request and retry) plus `_requestPlayPauseFocus` on `isRouteChainCurrent(context)`, the check `onWindowFocus` (`video_controls.dart:1187`) and the screen self-heal (`video_player_screen.dart:2012`) already carry. That closes every focus-claim path the player has under a covering route, not just the push instant.

Keep the rest as-is: the `restoreAfterResume` gate, the PiP latch through `_shouldSkipForPip` (mirrors the player's own skip decision, so the two can't disagree), and the `_remoteTransport` gate (needed on its own — the screen-level `Focus` derives `isCurrentRoute` from naive `isCurrent` at `video_player_screen.dart:2328`, so a media key landing on the screen node would otherwise resume the covered session).

Nit, non-blocking: `_activeState` duplicates `_activeRouteGuard._owner` bookkeeping at the same two sites; an `owner` getter on the guard would avoid the second static. A test in the `video_controls_select_key_test.dart` style — root picker over nested player, auto-hide elapses, focus stays on the picker — would lock the real fix in.

markwellis:
Changes pushed. I'm happy to test for a few days on my firetv to make sure there are no issues :)

markwellis:
this has been working well for me the last week.

I've rebased it on upstream, and removed one of the commits as it was implemented upstream

markwellis:
I've had no issues running this so far. Rebased on latest main branch, will continue to test - but it's ready for review/merge/feedback.

thanks!

markwellis:
hey @edde746 still running this build with no problems. I have rebased it against main today, no issues so far.

Is there anything you need further to consider it for merging? Thanks

Issue #2159, opened 2026-08-27
Home hubs: hide watched items, and hero style per hub


PilaScat:
### Problem Statement

Two things about the home screen hubs that no existing issue seems to cover.

**Watched titles never leave the rows.** Every hub keeps showing items I have already finished, so on a library I have worked through for years the genuinely new titles sit far to the right of rows I have to scroll past. Continue Watching already handles what is in progress; the other rows have no notion of "seen it".

**The hero style is all or nothing.** `showHeroSection` gives one hero at the top of the home screen, sourced from Continue Watching. There is no way to say "render *this* row as a hero and leave the rest as poster shelves", which is exactly what a curated collection deserves and what "Recently Added" does not.

Both of these used to be available through [replex](https://github.com/lostb1t/replex) (`REPLEX_EXCLUDE_WATCHED` and `REPLEX_HERO_ROWS`) as a proxy in front of Plex, but that project was archived in April 2025 and is looking for a maintainer, so it is no longer an option. Plezy renders these rows itself, so it is now the natural place for the behaviour.

### Proposed Solution

Two independent, opt-in settings under Appearance → Home Screen, next to the existing `useGlobalHubs` and `showServerNameOnHubs`. Both off by default, so nothing changes for anyone who does not go looking for them.

1. **Hide watched items in hubs** — filter fully watched items out of the home hubs. In-progress items stay, since Continue Watching already covers them.
2. **Hero-style rows, per hub** — let the user pick which hubs render as a full-width hero row with backdrop artwork instead of a poster shelf, rather than a single global switch.

Both can be applied client-side to the hubs already fetched in `lib/providers/discover_provider.dart`, so they should behave identically on Plex, Jellyfin and Emby.

### Alternatives Considered

Running replex as a proxy — no longer viable: archived, unmaintained, and Plex-only.

Related but different, so this is not a duplicate of any of them:

- #1652 (open) asks for merging home sections across libraries. Different axis: that one is about *which rows exist*, this one is about *what is inside a row* and *how a row is drawn*. I have commented there separately.
- #2066 asked for watched filtering on the **Discover** carousel and was closed as not planned. This is about the library hubs on the home screen, a different surface.
- #1998 asks to hide the watched **indicator**. That hides a badge; this removes the item from the row.
- #486 asked to change the source of the single top hero section and was closed as not planned. This is not about that hero: it is about letting an ordinary hub be drawn in hero style.

### Use Case

[...]

PR #2203, opened 2026-08-31
configurable home rows: merge libraries, collection rows (#1652)

### LINKED ISSUE #1652: Home Screen: Merge Sections Across Multiple Libraries
### Problem Statement

Feature Request: Customizable Home Screen Sections

First, I wanted to say thanks for the recent addition of genre tags. It was implemented quickly and looks great.

I have another idea that I think could make Plezy's home screen even more flexible.

Right now, each library gets its own "Recently Added" section. This works well for some setups, but users who organize Plex into many libraries can end up with a very long home screen.

It would be great if home sections could be configurable.

For example, instead of creating one row for every library, allow users to choose which libraries participate in a given section.

Example:

Recently Added (Merged)

- ✓ Movies
- ✓ New Releases
- ✓ TV Shows
- ✓ Cartoons
- ✗ Stand-Up Specials

This would create one combined "Recently Added" row using only the selected libraries.

The same library could also still have its own dedicated row if the user wanted both.

Examples:

- Recently Added (Merged)
- Recently Added Stand-Up Specials
- Recently Added Movies

I also think this concept could extend beyond "Recently Added" in the future, such as:

- Recently Released
- Continue Watching
- Random Picks
- Other home screen sections

The important part is giving users the ability to customize which libraries contribute to each section, rather than forcing a fixed layout.

This would let users tailor the home page to their own library organization while keeping the current behavior as the default for users who prefer separate rows.

Thanks again for all the work you've been putting into Plezy.

### Proposed Solution

I would like the ability to create customizable merged Home screen sections that combine content from multiple libraries into a single row.
For example, instead of separate rows for "Recently Added Movies," "Recently Added TV Shows," "Recently Added Cartoons," and "Recently Added New Releases," users could create a single "Recently Added" row that displays content from any libraries they choose.
Ideally, this would be fully customizable:
Enable or disable merged sections.
Choose exactly which libraries are included in each merged section.
Exclude specific libraries if desired.
Keep both the merged section and the individual library sections if preferred.
Expand the same concept to other Home screen sections in the future, such as Recently Released, Continue Watching, Recently Updated, or other customizable rows.
This would give users much more control over how their Home screen is organized, especially for servers with multiple libraries, while still preserving the current behavior for anyone who prefers separate sections.

### Alternatives Considered

The current workaround is to keep separate rows for each library, but this can make the Home screen feel cluttered on larger servers. Plex has a "Merge Recently Added Items" option, but Plezy could go even further by allowing users to customize exactly which libraries are merged and giving them the option to keep both merged and individual sections.

Image

### Use Case

1. As a user, I would like to create merged Home screen sections that combine content from multiple libraries while allowing me to choose exactly which libraries are included. I would also like the option to keep both the merged section and the individual library sections if desired.
2. This would help me organize the Home screen on larger Plex/Jellyfin servers. Instead of having separate "Recently Added" rows for Movies, TV Shows, Cartoons, New Releases, Stand-Up Specials, etc., I could create a single customized row while still leaving certain libraries separate if I choose.
3. The benefit would be a cleaner, more personalized Home screen with less clutter and greater flexibility. Every user could organize their Home page differently based on their library structure and preferences, making Plezy even more customizable than the official Plex client

### Target Platform

All platforms

### Media Server

Plex

### Additional Context

I included a screenshot of Plex's "Merge Recently Added Items" option as an example of the basic idea. My request is intentionally broader than Plex's implementation. The goal isn't just to merge "Recently Added," but to let users create customizable Home sections by choosing exactly which libraries are included or excluded for each row. This would make Plezy especially useful for larger servers with many libraries while still working well for smaller collections.

Related: #54 previously requested control over which libraries feed the Home page and their display order. This request builds on that idea by allowing selected libraries to contribute to customizable merged sections while optionally retaining their individual rows.

Image

jpdexter101-lang:
closes #1652.

adds configurable home rows:
- merge recently added / recently released across whichever libraries you pick. one combined row instead of one per library
- movie/show collection rows, scoped to whichever libraries or specific collections you pick. pick one collection and the row shows its titles directly, pick more and each shows as a poster you tap to open

settings > home layout:
- "add home row" to create one. pick the type, which libraries or collections feed it, and whether it also shows in library recommended and/or home
- new rows land at the top of the organizer, use the arrows to reorder
- tap a row in settings to edit it, delete removes it (asks to confirm)
- plex-managed rows (the ones from plex's own home manager) live in the same organizer. promote/demote with the "home" checkbox, remove entirely with the x

whatever order you set in the organizer is what shows on home now.

also added a line to the README's feature list for this.

tested: 81 tests, plus a manual pass adding/editing/deleting/reordering rows on a real windows build.

---

update: hero cards + trailer preview

any row (custom, plex-managed, or continue watching) can now be toggled to render as a hero card instead of a normal poster shelf, same row order, just bigger. on top of that, hero rows can optionally preview a trailer on hover:

- mouse hover dwells 3s before it plays, stops the instant you move off, goes back to the poster when the clip ends on its own
- continue watching hero cards preview your actual resume point instead of a trailer (capped at 30s), and clicking play still resumes from where you actually left off, not the preview's start
- hero/trailer settings follow the row wherever it's shown, not just home. a library's own recommended tab now respects the same per-row toggle
- tapping a hero card's poster and tapping it mid-trailer are separately configurable (play the item directly, or open its details page)
- continue watching gets its own per-library toggle + hero/trailer controls for that library's recommended tab, independent of the global home one
- reordering the organizer and each library's row list is click-hold-drag anywhere on the row now instead of arrow buttons, locked by default so a stray click can't reorder things
- continue watching's organizer entry is a checkbox now instead of an x, it had no way to get re-added once removed unlike every other row type

fixed along the way: the native video surface popping up in the wrong spot for a frame before snapping into place, letterboxed trailers showing black bars instead of filling their box, and a real bug where home and a library's recommended tab each had their own video controller and fought over the one video core the app actually has (showed up as a silently-black video on whichever one lost).

AI-assisted: Claude (Sonnet 5), used throughout for implementation, debugging, and testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Package managers:
 ### <img src="assets/readme_icons/browse.svg" height="20" alt="" align="center" /> Browse & Discover
 - Libraries, collections, and playlists — video and audio
 - Discover hub — Continue Watching, Next Up, trending, and recommendations
+- Configurable Home rows: merge Recently Added / Recently Released across whichever libraries you pick, or a collections row scoped to specific libraries or collections
 - Cross-server search across every connected Plex, Jellyfin, and Emby server
 - Filtering, sorting, and alphabetical jump navigation
 - Folder browsing and folder playback — home-video libraries open in folder view
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -95,6 +95,7 @@ import 'services/sensitive_prefs.dart';
 import 'services/startup_diagnostics.dart';
 import 'utils/dialogs.dart';
 import 'widgets/dialog_action_button.dart';
+import 'widgets/hover_preview/hover_preview_player_controller.dart';
 import 'widgets/startup_failure_view.dart';
 
 const bool _enableSentry = bool.fromEnvironment('ENABLE_SENTRY', defaultValue: false);
@@ -1600,6 +1601,18 @@ class _MainAppState extends State<MainApp> with WidgetsBindingObserver {
             return MultiServerProvider(_serverManager, _aggregationService);
           },
         ),
+        // App-wide singleton — Windows' native video plugin only backs one
+        // process-wide video core, so every hero row's hover-preview
+        // trailer, on any screen (Home, a library's own Recommended tab),
+        // has to share this one controller/Player. Two independently-owned
+        // controllers (one per screen) fight over that same native core the
+        // moment both are alive at once — even with only one on screen,
+        // Flutter's normal tab/route caching can keep the other mounted in
+        // the background — which showed up as a black, silently-failing
+        // video surface on whichever screen lost the race (reported
+        // 2026-09-10). Registered at the app root, not owned or disposed by
+        // any individual screen.
+        ChangeNotifierProvider<HoverPreviewPlayerController>(create: (_) => HoverPreviewPlayerController()),
         ChangeNotifierProxyProvider<MultiServerProvider, OfflineModeProvider>(
           create: (context) {
             final provider = OfflineModeProvider(
--- a/lib/media/media_hub.dart
+++ b/lib/media/media_hub.dart
@@ -31,6 +31,18 @@ class MediaHub {
   final String? serverId;
   final String? serverName;
 
+  /// Renders this row as a full-width rotating hero card instead of a
+  /// poster shelf. Resolved once when the hub is built (see
+  /// `home_section_builder.dart`) from either [HomeSectionConfig.heroStyle]
+  /// (custom rows) or the `managedHubHeroOverrides` setting (Plex-managed
+  /// rows and Continue Watching), so rendering code never has to re-derive
+  /// which store a given hub's hero flag lives in.
+  final bool heroStyle;
+
+  /// Plays a trailer/scene clip in the hero card. Meaningless without
+  /// [heroStyle] — same rule as the config types this is resolved from.
+  final bool heroTrailerPreview;
+
   const MediaHub({
     required this.id,
     required this.title,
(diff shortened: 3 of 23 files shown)


PilaScat:
Thanks for taking this to a PR. The two points I raised in #1652 are both still open in what's here, and I think both are fine as follow-ups rather than blockers:

- **Same-name collections across libraries.** Rows key off library/collection selection, so a Kometa list that exists as a movie collection *and* a show collection is still two picks producing two entries, rather than one merged ranked row. Matching by name is a different matcher from the one this PR needs, so a separate change.
- **Plex-only.** `plex_managed_hub.dart` and the `plex_client.dart` additions are Plex-specific, but `home_section_config.dart` and `home_section_builder.dart` read backend-neutral, so the Jellyfin/Emby follow-up looks like a per-backend hub fetch rather than a redesign.

The empty-merged-row bug you found (global/promoted hubs only returning promoted libraries) is a good catch — that one would have hit me directly, most of my libraries are not promoted.

One more thing, take it or leave it: if this lands, the per-row config in `home_section_config.dart` is also the natural place for the two options in #2159 (hero style per row, hide watched items), instead of the global switches they would have to be today.

jpdexter101-lang:
that last point about #2159, ended up doing exactly that in this update: hero style is a per-row toggle in home_section_config.dart now (checkbox next to the home/library checkboxes for every row, including continue watching), not a global switch. went further and did the trailer preview part of #2159 too: hover a hero card and it plays a preview instead of just sitting on the poster (a real trailer, or for continue watching your actual resume point, capped at 30s). same per-row setting also applies on a library's own recommended tab now, not just home.

didn't touch the same-name-collections-across-libraries merge or the jellyfin/emby side, those are still open like you said, not blockers for this.

Lokilicious:
@jpdexter101-lang If I may chime in, the hover gated trailer should ideally be a opt-in option for the user.

jpdexter101-lang:
It is

Issue #1804, opened 2026-08-05
Huge issues when playing Dolby Vision with TrueHD 7.1 on Android TV (2.12.1)


saakethkaashyap:
### Bug Description

On an armv7 Android TV device with a Realtek SoC, attempting to play 4K Dolby Vision MKV files (Profile 7 or Profile 8) with TrueHD Atmos 7.1 audio selected causes ExoPlayer playback to fail. 

When ExoPlayer drops the stream, it displays "Switching to compatible player..." and hands playback over to mpv, which then fails catastrophically:
- DV Profile 7 + TrueHD 7.1: App freezes or crashes completely.
- DV Profile 8 + TrueHD 7.1: Displays a solid green screen with no video playback.

Through testing, the issue was isolated to TrueHD Atmos 7.1 audio handling alongside Dolby Vision on this hardware. Switching the audio track to the compatibility track (AC3 5.1) completely fixes playback in ExoPlayer without triggering a fallback:
- Profile 7 plays natively via P7 -> P8.1 conversion.
- Profile 8 plays natively via DV P8 passthrough.

### Steps to Reproduce

1. Play a 4K Dolby Vision MKV file (DV Profile 7 or Profile 8) with TrueHD Atmos 7.1 audio selected on an armv7 Android TV.
2. Observe ExoPlayer failing and displaying "Switching to compatible player...".
3. Observe the fallback player (mpv) failing completely (freeze/crash on DV P7, solid green screen on DV P8).
4. Switch the audio track to AC3 5.1 and play again -> Observe video and audio playing perfectly natively in ExoPlayer.

### Expected Behavior

If TrueHD 7.1 audio passthrough or decoding is unsupported by the hardware pipeline, ExoPlayer should gracefully downmix/fallback the audio without failing the video renderer and triggering a broken mpv fallback. 

Same thing works well on KODI if used as external player

### Actual Behavior

Selecting TrueHD Atmos 7.1 audio alongside Dolby Vision video causes ExoPlayer to drop playback and hand off to mpv, which results in an app crash/freeze (DV Profile 7) or a solid green screen (DV Profile 8).

### Plezy Version

2.12.1
[plezy-android-armeabi-v7a.tar.gz](https://github.com/edde746/plezy/releases/download/2.12.1/plezy-android-armeabi-v7a.tar.gz)

### Platform

Android TV

### Media Server

Plex

### Device Information

[...]

PR #2231, opened 2026-09-03
fix(android): gate the DTS-HD IEC carrier on advertised DTS-HD support

JDTX0:
The ExoPlayer carrier gate only asked whether the route could open a 192kHz/7.1 IEC 61937 track. TrueHD rides that same tuple, so a sink that decodes Dolby but no DTS-HD passed the probe and was handed a DTS-HD burst it discards: the AudioTrack initialises and drains, Android reports no error, and neither the audio-recovery ladder nor the mpv fallback fires, so the file plays with no audio at all.

Evaluate the gate per format. TrueHD keeps the transport-only probe, so the #1804 routes are unchanged; DTS-HD additionally requires the route to advertise ENCODING_DTS_HD, the pairing mpvSpdifCodecs already applies and the reason mpv plays these files while ExoPlayer does not.

Declining the carrier is not a forced decode: the format falls through to media3's raw path, which downgrades DTS-HD to the DTS core for receivers that take the core but not the lossless stream.

My setup is a Fire TV Cube connected to an LG 4K OLED, with audio routed over eARC to a Sonos Arc Ultra. The Arc Ultra supports lossy DTS 5.1, but not DTS-HD MA or DTS:X.

Both affected MKV tracks are exposed to ExoPlayer as audio/vnd.dts.hd. Because the route supports the 192 kHz/7.1 IEC 61937 carrier used by TrueHD, Plezy incorrectly assumes it can also carry DTS-HD and sends the lossless DTS stream unchanged. The AudioTrack initializes successfully, but the Sonos cannot decode the stream, so playback has no audio. Android reports no error, preventing recovery or mpv fallback.

BTW, MPV does work correctly with DTS MA/DTS-X and I get audio output, but MPV stutters and drops frames on nearly all content I play. So I was forced to switch the default back to ExoPlayer, which triggered this particular issue.

Per the contribution guidelines: This code/PR was written by Claude Opus 5, high effort, and 1M context.

### DIFF
--- a/android/app/src/main/kotlin/com/edde746/plezy/exoplayer/AudioOutputPolicy.kt
+++ b/android/app/src/main/kotlin/com/edde746/plezy/exoplayer/AudioOutputPolicy.kt
@@ -163,10 +163,7 @@ internal fun mpvSpdifCodecs(
 @Suppress("DEPRECATION")
 @OptIn(UnstableApi::class)
 internal fun supportedMpvSpdifCodecs(context: Context): String {
-  val audioAttributes = AudioAttributes.Builder()
-    .setContentType(C.AUDIO_CONTENT_TYPE_MOVIE)
-    .setUsage(C.USAGE_MEDIA)
-    .build()
+  val audioAttributes = movieMedia3AudioAttributes()
   val capabilities = try {
     AudioCapabilities.getCapabilities(context, audioAttributes, null)
   } catch (error: Exception) {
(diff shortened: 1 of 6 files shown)


edde746:
> MPV stutters and drops frames on nearly all content I play

Could open an issue and get a log of this (with debug logging on)?

JDTX0:
@edde746 I've opened an issue for the MPV stuttering: https://github.com/edde746/plezy/issues/2240

If there's anything you need from me on this issue, or the other, let me know. I have access to ADB and can side-load to test if needed.

Issue #1128, opened 2026-05-25
no connection possible to server.


maxpowr1983-boop:
### Bug Description

plex client works fine. so the error is not server related. maybe plezy does not like my nginx reverse proxy?!

how to find out?

### Steps to Reproduce

plex client works fine. so the error is not server related. maybe plezy does not like my nginx reverse proxy?!

how to find out?

### Expected Behavior

plex client works fine. so the error is not server related. maybe plezy does not like my nginx reverse proxy?!

how to find out?

### Actual Behavior

plex client works fine. so the error is not server related. maybe plezy does not like my nginx reverse proxy?!

how to find out?

### Plezy Version

newest

### Platform

Windows

### Logs and Stack Traces

<details>
<summary>Logs</summary>

```
Paste logs here
```

</details>

edde746:
Logs?

maxpowr1983-boop:
where to get them on my windows machine? would provide them asap.

edde746:
Settings > View Logs

maxpowr1983-boop:
will get them.

maxpowr1983-boop:
9z9at

maxpowr1983-boop:
connects just fine on official client. but I want plezy FTW! :)

edde746:
Do you have some firewall or anti-virus that might be blocking Plezy from making connections?

maxpowr1983-boop:
no, plain win 11. defender deactivated. no anti virus, no firewall.

maxpowr1983-boop:
what does plex client do differently?

maxpowr1983-boop:
It connects fine on plezy using my android phone, same connection. so I guess it is the windows client here that is to blame.

edde746:
Could you add an exception for Plezy in the Windows firewall?

naddel91:
I deactivated the windows firewall just to test it. same results.

naddel91:
also tried giving plezy the right to go through the firewall. still no connection.

edde746:
Could you try [this build](https://github.com/edde746/plezy/actions/runs/26420998746/artifacts/7205421076) and send a new log?

maxpowr1983-boop:
qudui

naddel91:
just for comparison: here is the log on my other machine. same network, same plezy version, but it connects fine because it can load a cached connection?! 

4lymo

so there is something really off on the login-procedure. plex client works fine on both machines.

edde746:
[...]

PR #2233, opened 2026-09-03
fix(networking): follow the resolver's address ordering instead of dart:io's IPv4-first bias

lokok8:
## Symptom

I am using Plezy with a dualstack plex server. In the plex dashboard I was able to see that plezy prefers IPv4 over IPv6. The native plex player prefers IPv6 without issue. 

I decided to throw Claude at it and this change fixed it for me. Be aware that this was fully done by Claude other than me manually testing and verifying the produced build in my environment.

According to Claude diagnosis this only affects the API calls and not the actual media transfer.

I know everyone hates copy-pasted AI rambling, but hopefully the claude assession below is helpful.

## Root cause

`dart:io` does not connect in `getaddrinfo` order. `_NativeSocket.staggeredLookup`
issues the A-record lookup immediately and delays the AAAA lookup, and the SDK is
explicit about why:

```dart
// sdk/lib/_internal/vm/bin/socket_patch.dart
const concurrentLookupDelay = Duration(milliseconds: 10);

lookupAddresses(InternetAddressType.IPv4, ipv4Completer);
...
  // Introduce a delay before IPv6 lookup in order to favor IPv4.
  ipv6LookupDelay = Timer(concurrentLookupDelay,
      () => lookupAddresses(InternetAddressType.IPv6, ipv6Completer));
```

`tryConnectToResolvedAddresses` then consumes addresses in *arrival* order and only
releases the next candidate once the current one fails or its `_retryDuration`
(250 ms) expires. So on a dual-stack host whose IPv4 path answers in well under
250 ms, the AAAA address is never attempted at all — whatever RFC 6724 policy the
system resolver applied is discarded.

The system resolver's own ranking, for the same host on the same machine:

```
$ getent ahosts host.example
2001:db8::1    STREAM      <- RFC 6724 ranks IPv6 first
203.0.113.7    STREAM
```

Windows was never affected because it runs WinHTTP with
`WINHTTP_OPTION_IPV6_FAST_FALLBACK` (#1128). The doc comment in
`platform_http_client_io.dart` already noted that dart:io had no equivalent; this PR
supplies one for every other platform.

## The change

`lib/utils/happy_eyeballs.dart` (new) installs an `HttpClient.connectionFactory` that
does one `AF_UNSPEC` lookup via `InternetAddress.lookup` — preserving the resolver's
RFC 6724 ordering — interleaves candidates by family per RFC 8305 §4, and races them
with a 250 ms stagger (one constant, easily retuned), first connection winning and
the rest cancelled. The stagger
deliberately matches dart:io's own `_retryDuration` so the fallback is no slower than
the behaviour it replaces.

`_ConnectionTarget.connect` skips its own setup once a factory is set, so two
obligations transfer to us (both documented in the file):

- **TLS becomes ours.** The SDK calls `SecureSocket.startConnect` only when no factory
  is installed; with one set it wraps whatever socket it is handed in a plain
  `_HttpClientConnection`, so a raw socket returned for an `https` URL would speak
  cleartext to port 443. A test asserts a real handshake happens.
- **Proxied requests stay plain.** The SDK still drives `CONNECT` tunnelling and
  secures the tunnel itself, so the factory returns an unsecured socket to the proxy.

`HttpClient.badCertificateCallback` and a `SecurityContext` passed to
`HttpClient(context:)` are likewise bypassed by any connection factory. Plezy sets
neither today; the file documents it for whoever adds one.

### A second bypass, found only by running the app

`IOWebSocketChannel.connect` delegates to `WebSocket.connect`, which builds its **own**
private `HttpClient` unless handed one. The live-notification socket
(`/:/websockets/notifications`, 30 s ping) therefore stayed on the IPv4-first path even
with the factory installed — and as the app's longest-lived connection it is the one a
server is most likely to report as the client address. Fixed by passing a shared
factory-configured client as `customClient`; that is why the diff touches
`library_event_socket.dart`. Unit tests and offline measurement could not have found
it.

## Measurements

One host, one network — Fedora 44, Flutter 3.47.2 / Dart 3.13.2, dual-stack client,
server behind a custom HTTPS hostname. All figures are connect + TLS handshake.
Baseline is identical in every row because today's client always takes IPv4, so the
state of the IPv6 path is irrelevant to it: **~105 ms**.

| IPv6 condition | With this change | vs. today |
| --- | --- | --- |
| Healthy | **IPv6**, ~63 ms | ~40 ms *faster* — the v6 handshake won consistently |
| Dead at the routing layer (`ip -6 route add blackhole …`) | IPv4, ~102 ms | no change |
| SYNs silently dropped (`ip6tables -j DROP`) | IPv4, ~355 ms | **+250 ms** |

The middle row costs nothing because glibc applies RFC 6724 Rule 1 with no route to
the destination prefix and demotes the IPv6 address itself — `getent ahosts` flips to
IPv4-first under the blackhole route, so the ordering this PR follows is already
correct there.

## On the #1128 regression risk

The concern that motivated the WinHTTP swap is the mirror image of this one: a dead
IPv6 address stalling a connect. Two things bound it.

The stagger caps a black-holed candidate at one 250 ms interval rather than a connect
timeout — 250 ms out of the 2 s `MediaServerTimeouts.connectionRace` budget instead of
blowing through it. And it is not a new class of cost: dart:io already pays an
identical 250 ms whenever the *first* family it tries is the broken one, because
250 ms is its own `_retryDuration`. This change does not introduce the penalty, it
stops exempting IPv4 from it. The cost is per *connection*, not per request, so
`maxConnectionsPerHost = 12` with a 90 s `idleTimeout` amortises it across a
session.

## Verification

Live socket capture of the running Linux app — sockets attributed by `/proc/<pid>/exe`
and inodes matched against `/proc/net/tcp{,6}` — sampled every 20 s for two minutes:

```
before   IPv6=1..3   IPv4=1   <- pinned at 1 the whole window (the notifications socket)
after    IPv6=2..9   IPv4=0   <- zero for the whole window
```

Playback on the patched build: media stream on a single IPv6 socket, 264 MB
transferred, no IPv4 socket anywhere in the process.

**FFmpeg never had the bug and needs no change.** `libavformat/tcp.c` walks the
`getaddrinfo` list in resolver order, so it inherited the RFC 6724 ranking all along —
confirmed with system ffmpeg 8.1.2 (8/8 IPv6), a C probe against the pinned
`mpv-build` libmpv (4/4 IPv6), and the live stream above.

That sharpens the diagnosis: before this change Plezy talked to one server on **both
families at once** — media over IPv6 through FFmpeg, API traffic and the notifications
socket over IPv4 through dart:io. The inconsistency was internal to the app.

All four CI gates from CONTRIBUTING.md:

- `dart format --set-exit-if-changed` — clean
- `flutter analyze` — clean on every changed file (the project-wide issues are all
  pre-existing, in vendored `packages/wakelock_plus/{test,pigeons}`)
- `scripts/codegen.sh --check` — generated output current, tree unchanged
- `scripts/run_tests.sh` — **6580 passed, 6 skipped**, full suite

The 12 new tests cover resolver-order preservation, family interleaving, stagger
fallback, immediate release on a failed candidate, first-error reporting, cancellation
teardown, address-literal short-circuit, TLS-actually-happens, and proxy-stays-plain.

Built and ran `flutter build linux --debug` on Fedora 44 against a real Plex server.
`scripts/format_native.sh` does not apply (no native sources touched); the Maestro
Android E2E suites were not run.

## Follow-ups

Two further call sites share the `IOWebSocketChannel.connect`-without-`customClient`
shape and are not covered here:

- `companion_remote_peer_service.dart:83` and `:829` — same fix applies; left out
  because I could not exercise that subsystem to verify it.
- `watch_together_peer_service.dart:64` uses the cross-platform
  `WebSocketChannel.connect`, which has no `customClient` parameter; covering it means
  switching to `IOWebSocketChannel` on io platforms. It talks to the relay, not a
  user's media server.

### DIFF
--- a/lib/services/library_events/library_event_socket.dart
+++ b/lib/services/library_events/library_event_socket.dart
@@ -7,6 +7,7 @@ import 'package:web_socket_channel/web_socket_channel.dart';
 import '../../media/ids.dart';
 import '../../media/library_change_event.dart';
 import '../../utils/app_logger.dart';
+import '../../utils/happy_eyeballs.dart';
 
 /// Builds a live channel for [uri]. Injectable so tests can supply a fake or
 /// point at an ephemeral loopback server.
@@ -18,6 +19,9 @@ WebSocketChannel _defaultChannelFactory(Uri uri) => IOWebSocketChannel.connect(
   // Transport-level pings keep NAT mappings alive and surface a dead peer as
   // a close event; app-level keepalive (MediaBrowser) rides on top.
   pingInterval: const Duration(seconds: 30),
+  // Without an explicit client, WebSocket.connect builds its own and this
+  // channel bypasses the Happy Eyeballs connection factory entirely.
+  customClient: happyEyeballsHttpClient,
 );
 
 /// Reconnecting websocket base for one server's library-change push channel.
--- a/lib/utils/happy_eyeballs.dart
+++ b/lib/utils/happy_eyeballs.dart
@@ -1,0 +1,262 @@
+import 'dart:async';
+import 'dart:io';
+
+/// RFC 8305 connection setup for [HttpClient] that follows the platform
+/// resolver's address ordering instead of dart:io's IPv4-first bias.
+///
+/// `dart:io` does not connect in `getaddrinfo` order. `_NativeSocket
+/// .staggeredLookup` issues the A-record lookup immediately and delays the
+/// AAAA lookup by 10 ms — the SDK comment is literally "Introduce a delay
+/// before IPv6 lookup in order to favor IPv4" — and
+/// `tryConnectToResolvedAddresses` then consumes addresses in arrival order,
+/// releasing the next candidate only once the current one fails or its 250 ms
+/// stagger expires. On a dual-stack host whose IPv4 path answers in well under
+/// 250 ms the AAAA address is therefore never attempted at all, whatever
+/// RFC 6724 policy the system resolver applied.
+///
+/// That leaves Plezy the odd client out on a dual-stack network: browsers and
+/// the first-party Plex apps run Happy Eyeballs over the resolver's own
+/// ordering and land on IPv6, while Plezy pins the same server to IPv4.
+/// Windows never had the problem because WinHTTP supplies
+/// `WINHTTP_OPTION_IPV6_FAST_FALLBACK` (#1128).
+///
+/// This restores the platform's policy without giving up the fallback that
+/// motivated the Windows swap. [InternetAddress.lookup] returns the resolver's
+/// RFC 6724-sorted list, candidates are interleaved by family per RFC 8305 §4,
+/// and attempts are staggered, so a black-holed IPv6 address costs one
+/// [defaultAttemptDelay] rather than a connect timeout — 250 ms out of the 2 s
+/// `MediaServerTimeouts.connectionRace` budget, instead of blowing through it.
+
+/// Delay before the next candidate is raced alongside those already in flight.
+/// Matches dart:io's own `_retryDuration` so the fallback is no slower than
+/// the behaviour it replaces.
+const Duration defaultAttemptDelay = Duration(milliseconds: 250);
+
+/// Resolver seam. Defaults to [InternetAddress.lookup], whose result preserves
+/// `getaddrinfo` order.
+typedef AddressLookup = Future<List<InternetAddress>> Function(String host);
+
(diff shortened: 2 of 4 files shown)


edde746:
Closing in favor of https://github.com/edde746/plezy/commit/d371bfed3a1b154e945fd235faa5ce7054ec0ffd

Issue #2100, opened 2026-08-24
Live TV skip-back can seek forward after a time-shift reopen


nullroute77:
### Bug Description

When watching Plex Live TV with time-shifting, pressing the skip-back button can move the UI seek head backward while the actual video feed jumps **forward**.

I can reproduce this with either a 15-second or 30-second small-skip setting. This is not limited to the live edge: I paused long enough to build a large DVR buffer on both sides of the current playhead, and the same behavior still occurs.

Holding Left on the timeline for long enough does eventually rewind correctly.

### Steps to Reproduce

1. Start a Plex Live TV channel.
2. Pause long enough to build a substantial time-shift buffer.
3. Resume while well inside the seekable buffer, not at the live edge.
4. Press skip-back once (tested with both 15s and 30s).
5. Observe that the seek head moves backward, but after the Live TV stream reopens the actual video jumps forward.
6. Holding Left on the timeline long enough eventually does move playback backward.

### Expected Behavior

A relative skip-back should move the actual video backward by the configured Small Skip Duration, as long as that target is inside the capture buffer.

### Actual Behavior

The UI initially indicates a backward seek, but the reopened Plex Live TV stream can land later than the content that was playing before the seek

### Plezy Version

2.17.0

### Platform

Android TV, Windows

### Media Server

Plex

### Device Information

Shield TV Pro / Chromecast with Google TV 4K / Windows 11

### Additional Context

The following is a hypothesis based on the debug log and Plezy source produced by Codex.

Plezy Live TV relative seeks do not call a normal player seek; they rebuild/reopen the Plex HLS stream at an absolute DVR offset.

`LiveSeekAccumulator` appears to assume that after this reopen, the player's position will reset to approximately zero. It pins the requested target until `positionSeconds() < 2`, with a 1.5s ceiling.

In my debug log, the reopened mpv HLS streams instead report a non-zero `time-pos` of roughly 40–52 seconds after several time-shift reopens.

Plezy also sets the new stream epoch to the requested offset:

```dart
_live.streamStartEpoch = buffer.startedAt + offsetSeconds;
```

and later derives the absolute playback epoch as:

```dart
_rawPositionEpoch = _live.streamStartEpoch + player.state.position;
```

If mpv's `player.state.position` is already ~40–50 seconds after the reopen rather than ~0, this seems capable of double-counting that amount.

[...]

PR #2282, opened 2026-09-08
fix(livetv): start MPV offset streams at the first HLS segment

nullroute77:
Small Plex Live TV rewinds can reload into the same or later content. Plex has already positioned the replacement HLS playlist at the requested offset, but FFmpeg's default live startup selects the third-from-last available segment. MPV then rebases that starting point to zero, hiding the skipped content from the playback clock.

For native MPV opens with an explicit server offset, start at the first available segment using the file-local `demuxer-lavf-o-append=live_start_index=0` option. Watch-from-start and seek callers pass an explicit `timeShifted` flag, keeping backend-specific URL details out of the shared player code. Appending preserves other demuxer options, and file-local scope restores the defaults on subsequent opens. Dart regression coverage checks that the HLS option and quoted external subtitles are passed together; the native probe verified restoration on the next open.

Addresses the MPV rewind behavior reported in #2100. ExoPlayer remains unresolved and is outside this PR.

Validation:

- Manual testing: I tested Plex Live TV rewinding with MPV on an NVIDIA Shield TV Pro running Android TV and on Windows 11. The repair worked on both platforms.
- Native Windows probe against the pinned MPV DLL: default startup skipped 17 seconds in a synthetic live playlist; the first-segment option prevented the skip, preserved another demuxer option, and reset on the next open.
- Full Dart suite after merging current upstream: 7,087 passed, 6 skipped.
- Repository checks passed: Dart formatting, generated-code freshness, translation hygiene, hardcoded-string/workflow/script/icon guards, analyzer, and unused-code/file checks.
- Native formatting passed for Kotlin, Swift, and C/C++/Objective-C with Temurin JDK 21, ktlint 1.5.0, swift-format 6.3.3, and clang-format 18.1.3.

AI assistance: GPT-6 (Codex).


edde746 (changes requested):
Mechanism checks out: `-append` as a file-local `loadfile` option is the same path `_httpHeaderFieldsLoadfileOption` already uses, `sub-files` stays `%N%`-quoted so the join is safe, and the only global `demuxer-lavf-o` writer (VOD reconnect flags in `playback_open.dart`) survives the append and is restored on the next load. Live-edge reopens (retry, subtitle switch at edge, channel switch) never carry `offset`, so they keep FFmpeg's default, which is where they should be. Ran `test/mpv` and `test/screens/video_player` locally: green.

One change before merge:

**`live_tv.dart` — pass the flag explicitly instead of parsing `offset` out of the URL.** `_openLiveStream` is backend-neutral; `offset` is a Plex query key from `plex_client/parts/live_tv.dart`. Both callers that build a time-shifted URL already know it (`playback_start.dart` where `offsetSeconds != null`, and `_seekLivePosition`). Add `bool timeShifted = false` to `_openLiveStream`, pass `true` from those two sites, forward it as `startLivePlaylistFromBeginning: timeShifted`. That drops the `Uri.parse` per open and keeps the Plex wire detail out of the screen.

Nits, take or leave:

- The last two asserts in the new test (later live/VOD opens carry no option) are tautological at the Dart layer — the option string is a pure function of the call's arguments. The mpv-side restore is the real contract and your native probe is what covers it. The first assert (subtitles + flag in one arg vector) is the one worth keeping.
- `startLivePlaylistFromBeginning` is `PlayerNative`-only; worth saying "mpv" in the doc comment rather than describing it generically.

nullroute77:
Addressed the explicit timeshift flag request and both optional notes. All automated checks pass, and I’ve smoke-tested the updated build successfully. Ready for another review.

PR #2294, opened 2026-09-08
fix(emby): allow mpegts Live TV direct play

Kyomorie:
## Problem

Emby can report MPEG-TS Live TV sources with `Container: mpegts`, but Plezy's MediaBrowser direct-play profile currently only advertises `ts`.

That makes an otherwise directly playable Live TV source fail the container match during Emby playback negotiation. With Original quality selected, Plezy correctly enables DirectPlay and DirectStream, but Emby can still return an HLS transcode with:

`TranscodeReasons=ContainerNotSupported`

The same source is playable by Plezy's native player, and the Emby Android client can direct-play it on the same device/server.

This is separate from bitrate-based transcodes: this PR only fixes the `mpegts` container alias mismatch.

## Fix

Advertise both MPEG-TS spellings in the MediaBrowser video direct-play profile:

- `ts`
- `mpegts`

No playback policy, codec capability, bitrate handling, or transcoding profile is changed.

## Testing

Added a focused Emby Live TV regression test that:

- verifies the generated direct-play profile includes `mpegts`
- simulates Emby granting DirectPlay for a `Container: mpegts` live source
- verifies Plezy builds the static `/Videos/{id}/stream.mpegts` URL
- preserves `MediaSourceId` and `LiveStreamId`

Full repository CI is left to the pull-request checks.

## AI assistance

Implemented and reviewed with OpenAI GPT-5.6 Sol.


edde746:
I don't think the test is needed

Kyomorie:
> I don't think the test is needed

Makes sense — removed the dedicated test. Thanks 😊

PR #2328, opened 2026-09-11
feat(music): add discord rpc for music player

### LINKED ISSUE #2318: (Jellyfin) Add Discord RPC for Music
### Problem Statement

Discord RPC in Video Playback Settings doesn't work for Music playback, only traditional shows/movies. 

### Proposed Solution

Either do it from the same option, or ideally in a separate option for just Music playback (my ideal thing, as idc about showing off my shows).
I've already dug around the source-code to try to do this as a PR but I can't figure out where this would have to be added in (I'm not a Flutter dev lol)

### Alternatives Considered

_No response_

### Use Case

To show off my music listening habits on Discord, similar style to how Spotify's integration lets you do this.

### Target Platform

All platforms

### Media Server

Jellyfin

### Additional Context

_No response_

LivacoNew:
Adds Discord RPC when using the Music Player, via the same service as the Video RPC that's been slightly modified to allow for a Listening activity instead. Note this is toggled from the same setting as for Video RPC.

image

Tested on Linux using a Jellyfin server, not tried another media provider like Plex as I don't have it hosted myself.

Fixes #2318

### DIFF
--- a/lib/services/discord_rpc_service.dart
+++ b/lib/services/discord_rpc_service.dart
@@ -79,6 +79,7 @@ class DiscordRPCService {
   DateTime? _playbackStartTime;
   final PlaybackTimeline _timeline = PlaybackTimeline();
   double _playbackSpeed = 1.0;
+  DiscordActivityType _activityType = DiscordActivityType.watching;
   int _playbackRevision = 0;
   Timer? _reconnectTimer;
   DateTime? _lastPresenceUpdate;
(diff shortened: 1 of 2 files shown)


edde746:
Landed in 8e0891131 with a different wiring: presence is bound in `_bindTrackServices` alongside the tracker and OS media controls, pause/resume follow the player's playing flips, position ticks feed seek detection, and the player teardown clears presence. That last call is the reason `pausePlayback`/`resumePlayback` appeared to stick the card: the player subscriptions are cancelled before `player.stop()`, so no `playing=false` ever reached the handler. The service now derives the activity type from the item kind, and tracks use the album cover at a square size instead of `grandparentThumb` (the artist portrait on Plex). Ships in the next release.

LivacoNew:
Thanks for implementing this so quickly :)

> That last call is the reason pausePlayback/resumePlayback appeared to stick the card: the player subscriptions are cancelled before player.stop(), so no playing=false ever reached the handler

This seems to still be the case from testing off latest commit - Pausing my music will keep the embed up even when going back to the Home page. I still have to start a new stream just to make it go away.

image

^ Still there about 10 minutes after I had paused and stopped listening.

edde746:
That is pause, not stop: the session is still alive in the mini-player, and the card was deliberately kept while paused to match video. Two things made that wrong for music. A paused music session lives for hours, unlike a paused video screen. And Discord runs an "elapsed" counter from the activity's creation when no timestamps are sent, which is the 9:21 in your screenshot, so the card reads as still playing.

Fixed in 58773cb20: a paused track withdraws the Listening card and resume republishes it, same as Spotify's integration. Stop and the queue playing out clear it as well.

PR #2346, opened 2026-09-13
feat(subtitles): opt-in always-on subtitles and Plex auto-download

infixlabs:
Two opt-in settings under **Playback → Subtitles**, both off by default.

### Always Show Subtitles

Profile subtitle mode is deliberately not applied client-side for Plex — the comment above `_selectSubtitleTrackByProfile` explains why, and I have not touched that path. The consequence is that when PMS stamps no `selected` subtitle stream, the item plays unsubtitled and the account's "always" mode has no effect in the client.

This adds a separate local override below that. `_offOrForced` replaces the two off-returns a viewer who asked for always-on should override — the Plex "rows exist, none selected" branch and priority 5 — choosing preferred language, then any full-dialogue track, then a forced one. Full dialogue first, because a forced signs-and-songs track is not what "always show subtitles" means.

An explicit in-session off still wins: priority 1 returns before any of this, so turning subtitles off during playback is respected. `TrackSelectionService` gained two constructor parameters, both defaulted, so every existing call site and test is unchanged.

**I expect this to be the contentious part of the PR.** It is off by default and does not modify the intentional Plex behaviour, but it does sit on top of a decision the code documents as deliberate. Happy to rework or drop it if you would rather the client never second-guess PMS.

### Auto-Download Subtitles (Plex)

When the item carries nothing in the preferred language, this searches the server's providers and reuses the existing manual-download apply path (`_onSubtitleDownloaded`: poll `getVideoPlaybackData(forceRefresh: true)`, then switch source). One attempt per item per session, claimed *before* the network call so a throw cannot loop. Failures are silent apart from the log — it runs unprompted, often on a TV across the room.

Ranking puts `perfectMatch` first and score last. Score alone picked a confidently wrong subtitle for a film whose title is shared across releases, where only a file hash separates the right one; the full precedence is `perfectMatch` → requested language → full dialogue over forced/SDH → score.

### Smaller pieces

- Preferred language falls back to the app's own language rather than the system locale. The manual search sheet now shares that resolver, so the sheet and the new settings cannot disagree.
- `PlayerToastKind.busy` — the notice pill with a spinner — reports the fetch, since search + server-side download + poll runs for seconds over an unsubtitled picture. Success is silent; the subtitles are their own confirmation.
- `languageMatches` moved to a top-level `subtitleLanguageMatches` in `language_codes.dart` so the new pure helpers can use it. The instance method remains, delegating, so no call site changed.

### Testing

- `flutter analyze` clean, `scripts/run_tests.sh` 7208 passing
- New `test/services/subtitle_auto_download_test.dart` covers the ranking precedence (including the shared-title case) and the language matching
- `test/services/track_selection_service_test.dart` gains the forced-selection cases, including a regression guard that force-off still returns off at `serverSelected`

Tested on an Apple TV 4K against Plex.

---

AI-assisted: written with Claude Opus 5.


edde746:
Closing this — the feature conflicts with how subtitle selection is designed here.

The client deliberately treats the server as the single source of truth for automatic subtitle selection. Plex folds the account's `autoSelectSubtitle` into the `selected` stream flag, and "rows exist, none selected" is itself a server decision — the comment above `_selectSubtitleTrackByProfile` documents why the client must not re-apply the mode. `forceSubtitles` overrides exactly that decision and adds a second preferred-language setting that can disagree with the server's.

It doesn't stay local, either: a forced pick lands in `PlaybackSubtitleSelection`, carries across the episode boundary as a `SubtitleIntent`, and on the next episode matches at `navigation` priority — which persists it back to PMS via `selectStreams`. The "local" override silently becomes the item's server-side selection for every client.

Auto-download has the same problem in a stronger form: it writes a subtitle file into the server library unprompted, persists the stream selection when `rememberTrackSelections` is on, and overwrites an in-session explicit off (`rememberSourceSubtitlePreference` replaces `_sessionSubtitlePreference`). A background task shouldn't mutate shared server state or override a choice the viewer just made.

The `subtitleLanguageMatches` extraction and `pickBestSubtitleResult` ranking are fine pieces of code, but they exist to serve these two settings, so there's nothing to merge separately. If the underlying gap is "PMS doesn't apply the account's always mode," that fix belongs server-side, not as a client override.

PR #2349, opened 2026-09-14
feat(navigation): refresh the current page with Cmd+R or Ctrl+R on desktop

wthueb:
`gpt-5.6-sol` was used for assistance.


edde746:
Pushed two commits:

- `985ae5d` drops the `PageRefreshShortcut` widget — it had one consumer, and wrapping LogsScreen's build method re-indented ~90 lines for a three-line change. The chord is now handled in the screen's existing Focus handler, the pass-through helper is folded into `dispatchPageRefreshShortcut`, and `ManualRefreshable` is documented against `Refreshable`/`FullRefreshable`.
- `9d705e8` wires the search tab, where the chord previously did nothing while every other content tab refreshed.

Diff is now +237/-28 (was +380/-120).

PR #2360, opened 2026-09-15
fix(discord): hide rich presence while paused

Sekeyy:
Currently when paused, Discord Rich Presence shows "Watching Plezy" with a timer ticking for the paused state.

The current way for the paused state is completely fine and some people may prefer it, but an additional option I've seen used in Jellyfin rich presence implementations is an option for the activity to go away until playback starts again, which has become my preference and so I wanted to PR it as Plezy has become my main Jellyfin client.

I added this as an option under Discord Rich Presence and have tested it, the toggle seems to work completely fine after a restart, both on or off.

I tried to follow the guidelines presented through the Contributions markdown file, but if there is anything missing I would love to fix it so this can be implemented.


edde746:
Thanks for the PR — we took the behavior but dropped the setting.

A paused card is wrong rather than a preference: Discord runs an "elapsed" counter from the activity's creation when no timestamps are sent, so a paused card reads as still playing, and Rich Presence has no paused state to send. Music tracks already withdraw on pause for exactly this reason; this extends that to video. Gone on pause, back on resume — same convention as Spotify.

Pushed an amended commit (authorship preserved) that removes the pref/settings/i18n plumbing and makes hide-on-pause unconditional in `_updatePresence`. Note the original `setHideOnPause` had a bug — `_hideOnPause == hideToggle;` compared instead of assigning, so the toggle only took effect after a restart.

PR #2364, opened 2026-09-15
chore: apply verified maintenance updates

ShivamB25:
## Summary
- `sentry_flutter` (and the matched `sentry`) 9.25.0 → 9.30.0 for the Android JNI worker leak in
  [getsentry/sentry-dart#3987](https://github.com/getsentry/sentry-dart/pull/3987): a background worker outlived its
  Flutter engine, so its native/GPU allocations were never released. The fix stops a worker when its owning isolate
  exits and keeps native reporting across cached-engine view detach/reattach.
- SvelteKit 2.70.2 → 2.70.3 for the `$app/state` initialization fix (`FAQ.svelte` and `+error.svelte` import it).
- Drop the redundant `prerender` override on the privacy route; the root `+layout.ts` already exports
  `prerender = true`.
- Pinned `govulncheck` v1.1.4 → v1.8.0 to keep the scanner current with the module's Go 1.26 toolchain.

## What the Sentry bump actually carries

The range is four releases, not one PR. Two items matter beyond #3987:

- `sentry-android` 8.49.0 → 8.56.0.
- 9.28.0 ([#3972](https://github.com/getsentry/sentry-dart/pull/3972)) replaces the plugin's consumer rule
  `-keep class io.sentry.** { *; }` with a generated per-class allowlist. Audited: the allowlist is exactly the
  `classes:` list in the plugin's `ffi-jni.yaml` minus the three bootclasspath types (`java.net.Proxy`,
  `android.graphics.Bitmap`, `android.content.Context`), each entry carries `includedescriptorclasses`, and the
  `sentry-android-core` 8.56.0 AAR still ships `-keep class * extends io.sentry.SentryOptions { *; }`,
  `-keepnames class * implements io.sentry.Integration` and `-keeppackagenames io.sentry.**` for the JNI/NDK layer.
  Plezy itself has no `io.sentry` reference in Kotlin, Gradle or `proguard-rules.pro`, so nothing app-side depended on
  the old wildcard.

Nothing else in the range reaches this app: no `Sentry.logger`/`Sentry.metrics` use, so 9.28.0's
`enableLogs`/`enableMetrics` semantics change is inert; replay is off, so 9.27.0's replay network-detail capture is
inert; `tracesSampleRate = 0`, so 9.26.0's standalone app-start tracing is inert. `sentry-cocoa` stays pinned at
8.58.4 in both the 9.25.0 and 9.30.0 manifests, so the Apple SwiftPM locks are unchanged and
`check_apple_spm_locks.py` stays green.

## Maintainer edits

Rebased onto `main` and dropped commit `18798ee` (`refactor(android): remove disabled input and frame diagnostics`):

- `MainActivity.kt`'s `TEXT_INPUT_DIAGNOSTICS_ENABLED` is not permanently disabled. Commit `f4c83be`
  (`chore(tv): guard text input diagnostics`) introduced it together with `lib/utils/text_input_diagnostics.dart`, both
  logging under the same `TextInputDiag` prefix so one logcat capture interleaves the Dart key journey with native
  pre-IME/IMM state; the instrumentation came out of `ce9556d` (`fix(tv): restore the native Android IME for
  single-line text input`). The Dart half is still wired into `key_event_simulator.dart`, `focusable_text_field.dart`,
  `focusable_wrapper.dart` and `gamepad_service.dart`, so deleting only the native half leaves half a toggle. The
  deletion also left `isDpadKeyCode` with no callers, which no gate catches (no `allWarningsAsErrors`, ktlint is
  formatting-only).
- The `govulncheck` rationale was wrong: v1.1.4 reports
  `Govulncheck scanned the following 2 modules and the go1.26.6 standard library` and exits 0 on this module, so the
  gate was not broken. The bump is currency, and the commit message says that now.

## Validation on this branch

Run in a clean worktree at `main` + these four commits:

- `flutter pub get --enforce-lockfile` — resolves `sentry`/`sentry_flutter` 9.30.0
- `python3 scripts/checks/check_apple_spm_locks.py` — Apple SwiftPM locks match the resolved Sentry manifests
- `scripts/ci_checks.sh` — format, codegen freshness, translations, hardcoded strings, workflow/script guards, icons,
  native format, Dart analyzer, `dart_code_linter` unused code/files: all pass
- `flutter test` — 7,246 passed, 6 skipped
- `scripts/ci_server_checks.sh` — `go mod verify`, `go vet`, `go test -race`, `govulncheck@v1.8.0`: pass,
  no vulnerabilities
- `website`: `bun install --frozen-lockfile` (resolves kit 2.70.3, lock integrity matches the registry),
  `bun run check` 0 errors/0 warnings, `bun test` 17 pass/47 assertions, `bun run build` static build with
  `build/privacy.html` still prerendered

## Verification boundary

No Android device was attached, so the narrowed R8 keep rules were verified statically (allowlist vs `ffi-jni.yaml`,
AAR consumer rules) rather than by running a minified build. Before release, exercise
`python3 scripts/maestro/run_maestro_ci.py android-15-instrumentation` or a `minified` install and confirm a forced
crash still produces a Dart and a native event. No Apple or Windows runtime test was run either; the upstream PR
carries the device-level worker/engine lifecycle evidence.


edde746:
Pushed maintainer edits: rebased onto `main` and dropped `18798ee` (the Android diagnostics removal). The PR is now the two dependency bumps, the privacy-route cleanup, and the `govulncheck` pin, as four conventional commits. Full reasoning is in the updated description; the short version:

- `TEXT_INPUT_DIAGNOSTICS_ENABLED` is not permanently disabled. `f4c83be` added it alongside `lib/utils/text_input_diagnostics.dart`, both logging under the same `TextInputDiag` prefix so one logcat capture shows the Dart and native halves of a TV IME key press. The Dart half is still wired into four files, so removing only the native half leaves half a toggle. The removal also left `isDpadKeyCode` with no callers.
- The `govulncheck` justification was wrong: v1.1.4 scans the go1.26.6 standard library and exits 0 on `server/`. The bump is fine as currency; the commit message says that instead.

On the Sentry bump itself: the description now names what the four-release range carries, in particular 9.28.0 replacing the plugin's `-keep class io.sentry.**` consumer rule with a generated per-class allowlist. That surface is not covered by `check_shrinker_rules.py` (it scans `android/**` sources, not plugin rules) and `:app:testDebugUnitTest`/`lintDebug` never run R8, so please include a minified-build check in the future when a dependency changes keep rules. I verified it statically here (allowlist vs `ffi-jni.yaml`, plus the `sentry-android-core` 8.56.0 AAR rules); no device was attached for a runtime check.

Validation on the rebased branch: `scripts/ci_checks.sh` all pass, `flutter test` 7,246 passed / 6 skipped, `scripts/ci_server_checks.sh` pass with `govulncheck@v1.8.0`, website `bun run check` clean, `bun test` 17 pass, static build with `privacy.html` still prerendered, and `check_apple_spm_locks.py` green against the 9.30.0 manifests.

PR #2429, opened 2026-09-22
feat(episodes): show episode tags (canon/filler from Ronin) on season episode cards

### LINKED ISSUE #2428: feat(episodes): show episode tags (canon/filler from Ronin) on season episode cards
### Problem Statement

Anime plugins such as Ronin classify every episode of a show as Manga Canon, Anime Canon, Mixed Canon/Filler or Filler so viewers can skip filler. Ronin stores that classification as an ordinary Jellyfin item tag on each episode, but its badge only appears inside jellyfin-web, because the plugin injects it into the web client's page. Anyone watching through Plezy never sees it, so on a phone, desktop or TV you have to open the web UI or a filler list site to know whether the next episode matters.

Plezy already requests `Tags` on the item detail fetch and maps them onto its media model (they are visible in the metadata editor), but nothing renders them, and the season episode rows don't request the field at all.

### Proposed Solution

Show an episode's server tags on the season episode card, in the existing runtime / air date / quality row, as plain text labels. Control it with an Appearance setting, "Show Tags on Episodes", with three modes:

- **Canon and filler only** (default): only the four Ronin values (Manga Canon, Anime Canon, Mixed Canon/Filler, Filler), matched verbatim, so untagged libraries look exactly as before.
- **All tags**: every tag on the episode, as the server wrote it.
- **Off**.

Add `Tags` to the episode-row field set so the season list carries the tag without a second request. The server copies `Tags` off the item it has already loaded for the row, unlike the per-row count fields the field-set comments warn about, so the cost is negligible. No client-side processing beyond the optional filter: tags show exactly as the server wrote them.

### Alternatives Considered

- A Ronin-specific colored badge. Rejected because it ties the client to one plugin's look; plain labels serve any tag a server writes and still show the canon/filler mark.
- Reading the data from the plugin's own API. There is none. The tags are the data, and the item DTO Plezy already fetches is the only source needed.
- Renaming episodes with an "[F]" prefix on the server, as the AnimeFiller plugin does. Works in every client but mangles titles; a label beside the runtime keeps the title clean.
- Showing the tag on every surface (detail hero chips, TV metadata line, Continue Watching cards). Kept to the season episode card for a first pass, since that is where the skip decision is made. Other surfaces can follow.

### Use Case

1. As a Jellyfin user with Ronin installed, I open a show's season in Plezy and see "Filler" or "Mixed Canon/Filler" beside the runtime of the episodes it applies to, on phone, desktop and TV.
2. This lets me and everyone else on my server skip filler without leaving the app or checking a website.
3. The benefit extends to anyone who tags episodes on Jellyfin, Emby or Plex for any reason, since "All tags" shows whatever the server carries.

### Target Platform

All platforms

### Media Server

Jellyfin

### Additional Context

- Ronin: https://github.com/ahza-code/jellyfin-plugin-ronin. Its update task writes exactly one of "Manga Canon", "Anime Canon", "Mixed Canon/Filler", "Filler" into the episode's `Tags`, removing any previous one, and its web badge simply reads `Tags` off the episodes response with `Fields=Tags`.
- Jellyfin is the primary target. Emby returns the same tags as `TagItems` and Plex labels already map onto the same model field, so both work through the same code.
- In Plezy today, `Tags` is in the detail field set and mapped to the model's labels; the episode row field set stops at `MediaSources`.
- A PR implementing the above is ready and will be linked here once opened.

l3gitpanda:
Closes #2428

## Summary

Anime plugins such as [Ronin](https://github.com/ahza-code/jellyfin-plugin-ronin) classify each episode as Manga Canon, Anime Canon, Mixed Canon/Filler or Filler by writing an ordinary item tag onto it, and only jellyfin-web renders the mark. Plezy already maps `Tags` (Emby `TagItems`, Plex labels) onto `MediaItem.labels`; the season episode rows just never asked for the field and nothing rendered it.

- **Episode-row field set gains `Tags`** (`_baseEpisodeRowFields`). Unlike the count fields, the server copies it off the item it already loaded (`DtoService.AttachBasicFields`), so it is not another per-row query. Emby answers it with `TagItems`, which the mapper already reads.
- **New Appearance setting "Show Tags on Episodes"** (`SettingsService.episodeTags`, an `EnumPref<EpisodeTagsMode>`): Canon and filler only (default), All tags, Off. The default only ever shows the four Ronin values, matched verbatim, so untagged libraries look exactly as before; every other tag stays opt-in.
- **Episode cards** append the selected tags to the runtime / air date row as plain muted labels, ahead of the quality labels, on every platform. No client-side processing beyond the optional filter: tags show as the server wrote them.

## Tests

- `test/utils/episode_tag_labels_test.dart`: the filter for each mode, server order preserved, untagged episodes.
- `test/widgets/episode_card_test.dart`: the card shows the canon/filler tag by default, nothing in Off, everything in All tags, and reacts to the setting live.
- `test/services/jellyfin_client_urls_test.dart`: episode-row requests now assert `Tags` alongside `MediaSources`.
- `scripts/checks/check_analyzer.dart`, `dart format`, `scripts/codegen.sh --check`, the translation hygiene and icon checks all pass locally on Flutter 3.47.1.

## AI assistance

Implemented with Claude Code (Claude Fable 5.1).


edde746:
If you're making UI changes, attach screenshots.

l3gitpanda:
@edde746 Apologies, i was going to but i ended up passing out.

here is what it looks like, this is generally how id like it but id be fine with it looking different. The functionality would be great to have though!

image

PR #2432, opened 2026-09-23
fix(player): keep controls visible during held mouse moves

RyanTheTechMan:
Fix desktop player controls auto-hiding while the mouse moves with mouse button held.
Currently video player controls hide even while the user is holding down on any item, most noticeable when moving volume slider.

## Changes
- Route desktop mouse moves through the existing pointer-activity path.
- Keep mobile touch handling unchanged.
- Test held movement, normal hover, and auto-hide after movement stops.

AI-assisted using OpenAI GPT-6 Sol.

Video below shows before and after:
https://github.com/user-attachments/assets/002ce846-6ab7-4bdc-a36f-89a5122670be


edde746:
Superseded by #2436, which keeps this commit and adds a controls hold for the volume slider drag. Holding the slider still for 3s still hid the controls and dropped the new volume. The branch doesn't allow maintainer edits, so the follow-up couldn't be pushed here.

